PR #21, opened 2018-06-30
Assorted edits from today's review

jjhursey:


jjhursey:
I didn't get as far as I had hoped. but I'm going to release the lock on the standard until I get cycles again.

PR #23, opened 2018-08-10
Update the Introduction

rhc54:
Provide the couple of sections assigned to @rhc54


rhc54:
hmmm...travis isn't firing. Guess we lost/broke the connection?

PR #35, opened 2018-08-17
Add macros for supporting PMIx macros

rhc54:
Update Appendix on PRI extensions


rhc54:
@jjhursey I modified the appendix describing PRI extensions to only include the macro versions of the functions as I think we should push users in that direction.

PR #38, opened 2018-08-26
Fill in the Ack and History sections

rhc54:


rhc54:
@jjhursey I took a first pass a the Acknowledgements and History sections

Issue #77, opened 2018-09-13
mark Optional and PRI attributes as non-standard binding


abouteiller:
In many functions we take optional attributes, and we also expose the attributes that the PRI uses. In both cases, (and in contrast with the required attributes), these attributes are not required from an implementation.

Proposition is to 
1. reorder the attributes in all functions so that REQUIRED attributes are defined first
2. add to terms and definitions a new 'color box' similar to the advice to implementor type of box to denote optional and PRI attributes
3. Use the above term and definition space to clarify what we mean by optional and PRI defined attributes (i.e. optional are not required, but if provided must do what is depicted; PRI refer to the specific implementation of the reference server, and a rationale as to why we present them in the standard). 
4. clarify (at the same place) that in any case a compliant implementation does not have to support these attributes.

rhc54:
IIRC, we proposed to describe the _optional_ attributes in "advice to RM implementors" and drop the "optional" - i.e., we would say this is what a "quality implementation" would provide (without using that term - I forget what we agreed to use in its place).

I agree we need some clarifying text in the early part of the doc to explain why PRI is included and a clearer way of denoting those sections. We'll have to see if PRI-supported attributes are automatically considered _required_ or if some should have overlapping entries in the optional section.

PR #87, opened 2018-09-15
Define new commands for optional attributes

rhc54:
Add hyperlinks in red for PRI-supported attributes, and green for
PRRTE-supported attributes.

Refs #77 


rhc54:
@abouteiller @jjhursey What do you think of these markings? See PMIx_Init for what they look like.

jjhursey:
Having the note in section 2.5 about the difference between red and green is good. Though I only see red and blue. I don't see any use of the `\pastePRRTEAttributeItemBegin` in the current document.

rhc54:
I just added the first one - as you can imagine, they only really show up in the server areas. I'll be pushing it later.

PR #90, opened 2018-09-15
Typos and minor changes to event handling

abouteiller:
Signed-off-by: Aurélien Bouteiller <bouteill@icl.utk.edu>


Chap_API_Event.tex
```diff
@@ -41,13 +41,17 @@ \section{Notification and Management}
...
 
-Upon completing its work and prior to returning, each handler \textit{must} call the event handler completion function provided when it was invoked (including a status code plus any information to be passed to later handlers) so that the chain can continue being progressed. \ac{PMIx} automatically aggregates the status and any results of each handler (as provided in the completion callback) to that from all prior handlers so that each step in the chain has full knowledge of what preceded it. An event handler can terminate all further progress along the chain by passing the \refconst{PMIX_EVENT_ACTION_COMPLETE} status to the completion callback function.
+\adviceuserstart
+Note that \textit{first} and \textit{last} handlers are qualified within their category. That is, default event handlers will always be invoked after a single code event handler, even when the later is qualified as \textit{last}.
+\adviceuserend
```
rhc54: This is not actually correct - the first and last handlers are absolute and not within a category, and the last handler will be invoked after the default so it is truly the _last_ one.

rhc54: Maybe the correct question here is: we don't do what you describe today - should we? It would be a significant change to the current code, so it would be nice to first understand what is gained by doing it. Also need to clarify things like: is the "first" handler for the single code case different for each code? Or is it the "first" handler for all single-code registrations?

Could get awfully complex...

abouteiller: My bad. I misread the paragraph immediately above. Still a good thing that we review this piece.

I have no particular problem with the 'first' attribute being 'first' for all categories. 

I have a concern with the 'last' attribute. Given that it executes after the 'default' handler, and that the default handler often will abort the job, that could be of limited usefulness, or am I missing a point?

rhc54: My expectation was that the "abort" handler would be the one specified to execute last, but maybe that doesn't match what others would think. I can see where someone might feel that a "default" handler should play that role. Perhaps a "last" handler doesn't make sense?

rhc54:
Let me take this for now as I'm about to work on this chapter - I'll try to address the 'last' concern.

PR #100, opened 2018-09-24
Update to version 3 of the standard

rhc54:
Add a Security chapter. Add the few new APIs plus attributes. Show
attributes and error constants that are added in this version in magenta
as a way of highlighting them.

### DIFF
--- a/Chap_API_Security.tex
+++ b/Chap_API_Security.tex
@@ -1,0 +1,143 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% Chapter: Security
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\chapter{Security}
+\label{chap:api_security}
+
+Applications and tools often interact with each other in ways that require verification of the identity of the user making the request, and access to that user's relevant authorizations. This is particularly important when tools connect directly to a system-level \ac{PMIx} server that may be operating at a privileged level. A variety of system management software packages provide this service, but the lack of standardized interfaces makes portability problematic.
+
+This section defines two \ac{PMIx} client-side \acp{API} for this purpose. These are most likely to be used by user-space applications/tools, but are not restricted to that realm.
+
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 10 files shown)


pmix.sty
```diff
@@ -215,6 +215,7 @@
 \newcommand{\declareconstitem}[1]{\item[\code{#1}] \index{#1} \hspace{1em}}
 \newcommand{\declareconstitemvalue}[2]{\item[\code{#1}] \index{#1} \hspace{0.25em} \code{#2}  \hspace{1em}}
 \newcommand{\declareconstitemDEP}[2]{\item[\code{#1} (Deprecated in PMIx #2)] \index{#1} \hspace{1em}}
+\newcommand{\declareconstitemNEW}[1]{\item[\color{magenta}\code{#1}] \index{#1} \hspace{1em}}
```
jjhursey: I don't mind a 'new' highlight, which will move back to normal in the next release. Should we add the version marker (`\versionMarker{3.0}`) at the call locations to make sure they are labeled with the version they first appear in?

jjhursey:
I think this is good. Once we branch master for the `v2.0` standard, we can bring this in moving master to the `v3.0` standard.

pmix.sty
rhc54: I tried doing that first, but the marker is slightly shifted from the actual line and it gets confusing since each constant is basically one line. If we can get them better aligned, then that would indeed be an option.

rhc54:
Let's go ahead and commit. Any changes to v2 will be done on that branch and then brought forward to v3.

Chap_API_Security.tex
```diff
@@ -0,0 +1,143 @@
...
+
+Returns one of the following:
+
+\begin{itemize}
+    \item \refconst{PMIX_SUCCESS}, indicating that the request has been communicated to the local \ac{PMIx} server - result will be returned in the provided \refarg{cbfunc}
+    \item a \ac{PMIx} error constant indicating either an error in the input or that the request is unsupported - the \refarg{cbfunc} will \textit{not} be called
+\end{itemize}
+
+\reqattrstart
+\ac{PMIx} libraries are not required to directly support any attributes for this function. However, if the request cannot be processed by the library itself, then any attributes provided by the client must be passed to the host environment for processing. In addition, the following attributes are required to be included in the passed \refarg{info} array:
```
abouteiller: This is very unclear.

PR #101, opened 2018-09-27
Adjust spacing in TOC so labels don't overlap with numbers

jjhursey:
- Add a missing section header in server section.


jjhursey:
Here are some views from my build.
before
`^^ Before this PR`

after
`^^ After this PR`

Section number changed because I added a `\section` to the first block of `\subsection`'s

rhc54:
Kewl - thx! Go ahead and commit then

PR #105, opened 2018-10-10
Proof/v3x

abouteiller:


Chap_API_Init.tex
```diff
@@ -379,9 +379,9 @@ \subsection{\code{PMIx_tool_connect_to_server}}
 \adviceimplend
 
 \adviceuserstart
-Passing a \code{NULL} value for the \refarg{info} pointer is not allowed and will result in return of an error.
+Passing a \code{NULL} value for the \refarg{info} pointer is not allowed and will result in returning an error.
 
-\ac{PMIx} does not currently support on-the-fly changes to the tool's identifier. Thus, the new server must be under the same namespace manager (e.g., host \ac{RM}) as the prior server so that the original namespace assigned to the tool remains unique. The \refarg{proc} parameter is included here for obsolence protection in case this constraint is someday removed. Meantime, the parameter (if non-NULL) will be filled with the tool's existing nspace/rank.
+\ac{PMIx} does not currently support on-the-fly changes to the tool's identifier. Thus, the new server must be under the same namespace manager (e.g., host \ac{RM}) as the prior server so that the original namespace assigned to the tool remains unique. The \refarg{proc} parameter is included here for obsolence protection in case this constraint is someday removed. In the meantime, the parameter (if non-NULL) must be filled with the tool's existing nspace/rank.
```
rhc54: This last sentence confused me, I'm afraid. It could be interpreted that the caller must ensure that the parameter contains its name. Instead, the parameter is included so that the library must return the tools nspace/rank in that memory location. This is why I said it "will be filled" - as in, it "will be filled by the library". Maybe we should just copy the text from the ```PMIx_Init``` description for return of the proc parameter?

abouteiller: This text you are referring to I believe is more clear indeed. 
` returning the process identifier assigned to this client's application in the provided \refstruct{pmix_proc_t} struct.
Passing a value of \code{NULL} for this parameter is allowed if the user wishes solely to initialize the \ac{PMIx} system and does not require return of the identifier at that time.`

I'm not exactly sure why we need the comment about 'on-the-fly' changes then, but maybe I am missing something.

rhc54: It applies only to the case where the tool is disconnecting from its current server, and reconnecting to a different one. In that case, the tool could be assigned a new name by the new server. I don't know that we _need_ to support that case as you would think it only happens if the tool reconnects to a server in a totally different RM instance, which seems an extreme case. If we feel that is too extreme (after all, the tool could quit and restart to point somewhere else), then the comment can be removed.

abouteiller:
@rhc54 this has been updated to split  the 'on-the-fly' namespace into implementors and users parts advices

PR #123, opened 2018-10-27
Fix job-level information description

artpol84:
This PR clarifies namespace registration with respect to a job-level information format.

I was matching the way we interpret PMIx standard in Slurm and I found it difficult to implement based on description only. I had to go consult PRRTE implementation for that. 
The main thing is that it seems to be incomplete with respect to how info argument should look like.

In particular:
* From the description it seems like job-level information has a flat organization. While in practice Process data for the particular process is grouped in the `PMIX_PROC_DATA` key.
* Application notion is also confusing there. As we don't have an application number argument in PMIx_Get there is no way to request keys about applications. Because of that, there is no way for the client to request information about "given application". As the result, PRRTE is storing application number, application size and other stuff for each process. However from the description it seems that RM is expected to group info about application (number, size, appldr) at higher level.

I tried to modify the standard to make it more straightforward.

### DIFF
--- a/Chap_API_Server.tex
+++ b/Chap_API_Server.tex
@@ -160,15 +160,12 @@ Host environments are \textit{required} to provide the following attributes:
 If more than one application is included in the namespace, then the host environment is also \textit{required} to provide the following attributes:
 
 \begin{itemize}
-    \item for each application:
+    \item for each process:
         \begin{itemize}
             \item \pastePRRTEAttributeItem{PMIX_APPNUM}
             \item \pastePRRTEAttributeItem{PMIX_APPLDR}
             \item \pastePRRTEAttributeItem{PMIX_APP_SIZE}
-        \end{itemize}
-    \item for each process:
-        \begin{itemize}
-            \item \pastePRRTEAttributeItem{PMIX_APP_RANK}
+             \item \pastePRRTEAttributeItem{PMIX_APP_RANK}
         \end{itemize}
 \end{itemize}
 
@@ -185,17 +182,12 @@ The following attributes \textit{may} be provided by host environments:
             \item \pastePRRTEAttributeItem{PMIX_SERVER_NSPACE}
             \item \pastePRRTEAttributeItem{PMIX_SERVER_RANK}
             \item \pastePRRTEAttributeItem{PMIX_NPROC_OFFSET}
-            \item \pastePRRTEAttributeItem{PMIX_APPLDR}
             \item \pastePRRTEAttributeItem{PMIX_SESSION_ID}
             \item \pastePRRTEAttributeItem{PMIX_ALLOCATED_NODELIST}
             \item \pastePRRTEAttributeItem{PMIX_JOB_NUM_APPS}
             \item \pastePRRTEAttributeItem{PMIX_MAPBY}
             \item \pastePRRTEAttributeItem{PMIX_RANKBY}
             \item \pastePRRTEAttributeItem{PMIX_BINDTO}
-        \end{itemize}
-    \item for each application in the given namespace:
-        \begin{itemize}
-            \item \pastePRRTEAttributeItem{PMIX_APP_SIZE}
         \end{itemize}
     \item for its own node:
         \begin{itemize}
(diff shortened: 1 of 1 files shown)


Chap_API_Server.tex
```diff
@@ -160,15 +160,12 @@ \subsection{\code{PMIx_server_register_nspace}}
 If more than one application is included in the namespace, then the host environment is also \textit{required} to provide the following attributes:
 
 \begin{itemize}
-    \item for each application:
```
artpol84: As mentioned in the PR header there is no such entity in PMIx as an application. 
For example PRRTE is storing this data on per-process basis:
https://github.com/pmix/prrte/blob/master/orte/orted/pmix/pmix_server_register_fns.c#L418:L448
This separation seems misleading for me.

Chap_API_Server.tex
```diff
@@ -185,18 +182,13 @@ \subsection{\code{PMIx_server_register_nspace}}
             \item \pastePRRTEAttributeItem{PMIX_SERVER_NSPACE}
             \item \pastePRRTEAttributeItem{PMIX_SERVER_RANK}
             \item \pastePRRTEAttributeItem{PMIX_NPROC_OFFSET}
-            \item \pastePRRTEAttributeItem{PMIX_APPLDR}
```
artpol84: APPLDR was already mentioned in the `required` section

Chap_API_Server.tex
```diff
@@ -185,18 +182,13 @@ \subsection{\code{PMIx_server_register_nspace}}
...
             \item \pastePRRTEAttributeItem{PMIX_SESSION_ID}
             \item \pastePRRTEAttributeItem{PMIX_ALLOCATED_NODELIST}
             \item \pastePRRTEAttributeItem{PMIX_JOB_NUM_APPS}
             \item \pastePRRTEAttributeItem{PMIX_MAPBY}
             \item \pastePRRTEAttributeItem{PMIX_RANKBY}
             \item \pastePRRTEAttributeItem{PMIX_BINDTO}
         \end{itemize}
-    \item for each application in the given namespace:
-        \begin{itemize}
-            \item \pastePRRTEAttributeItem{PMIX_APP_SIZE}
```
artpol84: `app_size` was also mentioned in the required section.

artpol84: Now thinking of that I figured out that it might be related to the case of single-application where RM is still allowed to publish this info key.
If this is the case I can revert this change.

Chap_API_Server.tex
artpol84: APPLDR seems to belong to the application-related section

Chap_API_Server.tex
rhc54: Oh my - I really disagree. There most certainly is the concept of an application in PMIx. I agree that PRRTE may be storing things differently than described here, but that may be an issue with PRRTE or could be an error here. I'd have to look more carefully.

Remember, we are doing this after-the-fact. We already had to make adjustments to the PMIx library to comply with the standard as it evolved - not reason to believe that similar changes to PRRTE and/or the Slurm plugin will be required.

Chap_API_Server.tex
rhc54: Yes, APPLDR represents the global rank of the first process of the given application. I think the confusion may perhaps be that we may not have used "job" and "application" separately enough - i.e., we may have interchanged them in some places. We can certainly read thru and ensure we are more careful about it.

Chap_API_Server.tex
rhc54: I should think "app_size" would be required.

Chap_API_Server.tex
```diff
@@ -223,6 +215,37 @@ \subsection{\code{PMIx_server_register_nspace}}
...
+        \begin{itemize}
+            \item required: \refattr{PMIX_APPNUM}, \refattr{PMIX_APPLDR}, \refattr{PMIX_APP_SIZE}, \refattr{PMIX_APP_RANK}
+        \end{itemize}
+        \item process-level key-values:
+        \begin{itemize}
+            \item required: \refattr{PMIX_RANK}, \refattr{PMIX_LOCAL_RANK}, \refattr{PMIX_NODE_RANK}, \refattr{PMIX_NODEID}
+            \item optional: \refattr{PMIX_PROCID}, \refattr{PMIX_GLOBAL_RANK}, \refattr{PMIX_HOSTNAME}
+        \end{itemize}
+    \end{itemize}
+\end{itemize}
```
rhc54: I'm not sure I understand how this differs from the prior itemized list of required and optional values. We certainly don't want to list them all twice, so perhaps we just need to look at organization?

rhc54:
How odd - GitHub posted my review comments twice! Sorry about that - I have no idea why it did it.

Chap_API_Server.tex
artpol84: > Oh my - I really disagree. There most certainly is the concept of an application in PMIx.
There is a mention of the application in PMIx, but no program object in API that corresponds to it. At least the way I read the standard.
For example, how can client request information about the size of applicatoin number 2?

It can reference jobs (by namespaces) so can request the size of job corresponding to namespace "ns2" by calling
```C
PMIx_Get("ns2", PMIX_RANK_WILDCARD, PMIX_JOB_SIZE, ...);
```
It can ask for some process-specific information, for example for process 2 it can call
```
PMIx_Get("ns1", 2, PMIX_NODEID, ...);
```
But how will it reference specific application number?

Chap_API_Server.tex
artpol84: From my standpoint it is more than that, see my comment below about how to access the key of a specific application.

Chap_API_Server.tex
artpol84: Meaning you agree that we can remove it here?
But what if it is a single application job. According to the standard RM is not required to post application keys, but I assume it still can do it if it wants to.

Chap_API_Server.tex
artpol84: I'd personally prefer to have a summary of the keys and the groups they belong to.
The original description is less visual from my perspective also given that it also provides meanings of each key.
IMO grouping by key classes (job-,node-,app-,proc-level) and then by required/optional would improve interpretation. Because this is what implementor will need to do - group the keys.

I also think that we need to explicitly say that process information has to be grouped. Otherwise PMIX_PROC_DATA key is meaningless in the standard. It is defined but never used.

I recognize that other people may think in different way so maybe it is a good thing to provide both classifications.

rhc54: I'm afraid we already decided quite a while back to group attributes by "required" vs "optional", and it is too late to revisit that decision. What we can do is group things within those categories, and I'm happy to take a crack at it.

Chap_API_Server.tex
rhc54: Meaning that app_size is required, but I need to look at the total section and not just these commit snippets to see how to organize it. 😄

Chap_API_Server.tex
rhc54: You use the info keys - I can add an example to the wording. We use it in various places today.

Chap_API_Server.tex
artpol84: > What we can do is group things within those categories, and I'm happy to take a crack at it.

That may be ok, but my original point was to have just a key list without explanations to provide a summary. If we will re-group it won't be a summary.

Chap_API_Server.tex
artpol84: Here I think I was deleting app_size from the "optional" subgroup.

Chap_API_Server.tex
rhc54: Well, like I said, we already decided on a format - and we really need to stick with it. We have so far gotten very positive reviews on it 😄

artpol84: But I'm keeping the original grouping providing an additional summary. I don't think it confronts the original format.

Chap_API_Server.tex
artpol84: Then I guess application-level information should be grouped the same way proc-level keys are being grouped today.
But I don't see PMIX_APP_DATA key for that.

artpol84: I mean grouped by host RM when submitting it

Chap_API_Server.tex
artpol84: > Meaning that app_size is required, but I need to look at the total section and not just these commit snippets to see how to organize it. 

You have one negative from me now.

Chap_API_Server.tex
artpol84: > Well, like I said, we already decided on a format - and we really need to stick with it. We have so far gotten very positive reviews on it 😄

One negative from me

artpol84: Though I agree that this grouping helps a bunch, but for the cases where there is a lot of possible parameters it is a good idea to provide a summary. And that's what I did.

Chap_API_Server.tex
rhc54: I'll try to clarify the text and provide an example or two.

Chap_API_Server.tex
rhc54: Understood - sorry you don't like the format, but it was decided upon by those participating at the time. We can't change it now, but we can try to provide some clarifications in the hope it meets the general needs.

artpol84: You don't quite answer my question.
I'm not arguing about changing the format. Only proposing to provide additional clarification. Why is that not appropriate?

artpol84:
A more generic solution is provided in #126

PR #151, opened 2019-01-24
Clarify that regex preserves ordering

jjhursey:


Chap_API_Server.tex
```diff
@@ -44,6 +44,7 @@ \subsection{\code{PMIx_generate_regex}}
 \descr
 
 Given a comma-separated list of \refarg{input} values, generate a regular expression that can be passed down to the \ac{PMIx} client for parsing.
+The order of the hosts in the input string is preserved in the regex.
```
abouteiller: Is the generate_regex specific for host lists or generic? It is the first time that the function is tied to the concept of host. I would say that the regex preserve ordering, in general terms.

jjhursey: Good point. I'll adjust the text

jjhursey: I just pushed a re-wording. Let me know what you think.

abouteiller: 👍

PR #156, opened 2019-01-25
Report supported attributes for specified functions

rhc54:
Provide a mechanism by which users can request a list of the supported
attributes for a given function. Add a new API by which the host
environment can register their support - the PMIx library will register
its own supported attributes internally during init.


rhc54:
I agree, but I also couldn't find an alternative way of doing it - especially under the requirement that you not have to call the API itself before you can query its support. However, the standard doesn't stipulate _how_ one does it, so if someone does come up with a better idea, we aren't preventing them from pursuing it.

Chap_API_Struct.tex
```diff
@@ -2151,9 +2151,118 @@ \subsubsection{Create the\refstruct{pmix_info_t} array of query qualifiers}
...
+\begin{codepar}
+typedef struct pmix_regattr \{
+    char *key;
+    pmix_data_type_t type;
+    char **valid_values;
+\} pmix_regattr_t;;
+\end{codepar}
+\cspecificend
+
+Characters in the key must be standard alphanumeric values supported by common utilities such as \textit{strcmp}. The \refarg{type} must be a \ac{PMIx} data type identifying the type of data associated with the \refarg{key} attribute. The \refarg{valid_values} field consists of a \code{NULL}-terminated array of strings identifying the valid values (or range of values) plus any other desired explanation.
```
jjhursey: The `plus any other desired explanation` part is somewhat confusing for me. The use case I was thinking of is having a list of valid values that I can then turn around and use. But if that valid value has concatenated with it a description then I cannot do that without knowing the boundary.

I'm thinking that adding an explanation in that field is not preferred, and instead of having a separate field for documentation or leaving it off and requiring the RM to document any restrictions. Programmatically a description like that isn't useful except for printing back to the user.

Chap_API_Struct.tex
```diff
@@ -2151,9 +2151,118 @@ \subsubsection{Create the\refstruct{pmix_info_t} array of query qualifiers}
...
+\subsection{Attribute registration structure}
+\declarestruct{pmix_regattr_t}
+
+The \refstruct{pmix_regattr_t} structure is used to register attribute support for a \ac{PMIx} function.
+
+\versionMarker{4.0}
+\cspecificstart
+\begin{codepar}
+typedef struct pmix_regattr \{
+    char *key;
```
jjhursey: It's still a little unclear to me if the `key` is the macro constant or the string constant behind the macro. Meaning, for example, if we have this attribute:
```
PMIX_SERVER_TMPDIR "pmix.srvr.tmpdir"
```
Should `key = strdup("PMIX_SERVER_TMPDIR");` or should `key = strdup(PMIX_SERVER_TMPDIR)` (where the latter duplicates `pmix.srvr.tmpdir`

I can see an argument for both. I guess the question is do we expect a user to do a `strcmp(key, "PMIX_SERVER_TMPDIR")` or a `strcmp(key, PMIX_SERVER_TMPDIR)` in thier program.

Chap_API_Struct.tex
rhc54: I think there are two use-cases here:

1. request from a tool (e.g., pattr) that displays the result to a human user. In this case, we want to be able to provide any additional description

1. request from a program that intends to parse the result. In this case, we want to clearly separate out and describe the accepted values in a machine-understandable way.

What if we add an attribute that requests the output be in machine-parsable form? This would require something other than just strings for valid values - e.g., the library currently returns "RANGE OF POSITIVE INTEGERS" for at least one attribute valid value, but would we want to declare that response as part of the standard? Or is it better to have attributes for minimum and maximum values that can include the data type (i.e., float vs int vs uint16_t)?

In other words, if we want this to be machine-parsable, then we have to be more descriptive of the valid values. Do we want to do that, or just support human-readable output?

Chap_API_Struct.tex
rhc54: It isn't that much more work to simply accept both, frankly - i.e., we can say that the PMIx library must accept either form. The bigger question is: what should be returned?

I think the human-readable output should be the attribute name - i.e., "PMIX_SERVER"TMPDIR".

The machine-parsable output should be the attribute representation - i.e., "pmix.srvr.tmpdir" - as that is what the program would need to pass back to a PMIx API.

Make sense?

rhc54: Actually, let me correct myself - the input is always the _function_ whose attribute support is being requested, never an attribute itself. So this really just concerns the two options for output.

Chap_API_Struct.tex
jjhursey: The more I think about it the more this sounds like the discussion around MCA parameters in OMPI. Originally available as strings to display via `ompi_info`, but later some were exposed more programatically for tools to play with.

Maybe for this first pass we focus on a `ompi_info` like tool to report out to the user, then think more about the latter case as an extension of this functionality.

With that in mind, I think having the `valid_values` be an array of descriptive strings is probably fine. If there was a way to express the separation between the value and the description, when necessary, I think that would be helpful. For example, something like:
```
valid_values[0] = strdup("values < -1 are reserved for future use");
valid_values[1] = strdup("\"-1\" indicates enable X with default values");
valid_values[2] = strdup("\"0\" indicates disable X");
valid_values[3] = strdup("values > 0 indicates enable X with that specific value");
```
So the `-1` and `0` cases are identified as specific values by using quotes (or some other unique character), but the rest express a general comment for the user about ranges of values.


Having an interface that can be inspected more programatically would be helpful if I were writing a program that wants to adapt to the availability of different attributes/values. But I think I'd want someone intending to do that to give some guidance on how to form up the interface. Maybe post this as an issue for a separate discussion.

Chap_API_Struct.tex
jjhursey: On the server side, we would need an ability to translate the attribute representation string into the attribute name macro string. Not too hard to do. So we could accept either. For RM custom attributes we might only have the attribute representation to pass through the interface.

Considering the possibility of RM custom attributes being expressed through this interface, I think on the client/tool side we should try to report both the attribute name (if a translation is available) and the attribute representation. The latter will always have to exist, but the former may not for RM custom attributes.

rhc54:
@jjhursey Please take a look at the more recent changes. I tried to address both of your concerns.

Chap_API_Struct.tex
```diff
@@ -2151,9 +2151,142 @@ \subsubsection{Create the\refstruct{pmix_info_t} array of query qualifiers}
...
+    \item the \refarg{string} is the literal string value of the attribute - e.g., "pmix.max.size" for the \refattr{PMIX_MAX_PROCS} attribute
+    \item \refarg{type} must be a \ac{PMIx} data type identifying the type of data associated with this attribute.
+    \item the \refarg{info} array contains machine-usable information regarding the range of accepted values. This may include entries for \refattr{PMIX_MIN_VALUE}, \refattr{PMIX_MAX_VALUE}, \refattr{PMIX_ENUM_VALUE}, or a combination of them. For example, an attribute that supports all positive integers might delineate it by including a \refstruct{pmix_info_t} with a key of \refattr{PMIX_MIN_VALUE}, type of \refconst{PMIX_INT}, and value of zero. The lack of an entry for \refattr{PMIX_MAX_VALUE} indicates that there is no ceiling to the range of accepted values.
+    \item The \refarg{description} field consists of a \code{NULL}-terminated array of strings describing the attribute, optionally including a human-readable description of the range of accepted values - e.g., "ALL POSITIVE INTEGERS", or a comma-delimited list of enum value names.
```
jjhursey: Missing `\item` for `ninfo`

rhc54:
@jjhursey Added the requested tweaks - see if this addresses your questions.

PR #160, opened 2019-01-31
Clarify some things per filed issues

rhc54:
From updated history:

* Remove PMIX_NUM_SLOTS as it is duplicative of (a) PMIX_UNIV_SIZE when used at the session level and (b) PMIX_MAX_PROCS when used at the job and application levels

* Clarify difference between PMIX_JOB_SIZE and PMIX_MAX_PROCS

* Clarify that PMIx_server_setup_application must be called per-job instead of per-application as the name implies. Unfortunately, this is a historical artifact. Note that both PMIX_NODE_MAP and PMIX_PROC_MAP must be included as input in the info array provided to that function. Further descriptive explanation of the ``instant on'' procedure will be provided in the next version of the PMIx Standard.

* Clarify how the PMIx server expects data passed to the host by pmix_server_fencenb_fn_t should be aggregated across nodes, and provide a code snippet example


Chap_Introduction.tex
```diff
@@ -389,5 +389,9 @@ \section{Version 3.1: Jan. 2019}
     \item Add new attributes to specify the level of information being requested where ambiguity may exist (see \ref{api:struct:attributes:retrieval})
     \item Add new attributes to assemble information by its level for storage where ambiguity may exist (see \ref{api:struct:attributes:storage})
     \item Add flag and \refmacro{PMIX_INFO_IS_END} macro for marking and detecting the end of a \refstruct{pmix_info_t} array
+    \item Remove \code{PMIX_NUM_SLOTS} as it is duplicative of (a) \refattr{PMIX_UNIV_SIZE} when used at the \refterm{session} level and (b) \refattr{PMIX_MAX_PROCS} when used at the \refterm{job} and \refterm{application} levels
```
abouteiller: Isn't that a non-backward compatible change (although it is ABI compatible)?

rhc54: Fair point - suggestions? As I see it, we could:

- leave the duplication in and just add the explanation to the definition
- handle the conversion to/from PMIX_NUM_SLOTS in the library
- ??

The first option is obviously the easiest, but is that what we want to do?

Chap_API_Server.tex
```diff
@@ -779,13 +779,13 @@ \subsection{\code{PMIx_server_setup_application}}
...
+Host environments are \textit{required} to execute this operation prior to launching a job. In addition to supported directives, the \refarg{info} array \emph{must} include a description of the \refterm{job} using the \refattr{PMIX_NODE_MAP} and \refattr{PMIX_PROC_MAP} attributes.
 \advicermend
 
-This is defined as a non-blocking operation in case contributing subsystems need to perform some potentially time consuming action (e.g., query a remote service) before responding. The returned data must be distributed by the \ac{RM} and subsequently delivered to the local \ac{PMIx} server on each node where application processes will execute, prior to initiating execution of those processes.
+This is defined as a non-blocking operation in case contributing subsystems need to perform some potentially time consuming action (e.g., query a remote service) before responding. The returned data \emph{must} be distributed by the \ac{RM} and subsequently delivered to the local \ac{PMIx} server on each node where application processes will execute, prior to initiating execution of those processes.
```
gvallee: It seems that we do not emphasis *must* in the rest of the document so we should not do it here.

Chap_API_Server.tex
```diff
@@ -1300,7 +1300,21 @@ \subsection{\code{pmix_server_fencenb_fn_t}}
 \adviceimplend
 
 \advicermstart
-The host will receive a single call for each collective operation. It is the responsibility of the host to identify the nodes containing participating processes, execute the collective across all participating nodes, and notify the local \ac{PMIx} server library upon completion of the global collective.
+The host will receive a single call for each collective operation. It is the responsibility of the host to identify the nodes containing participating processes, execute the collective across all participating nodes, and notify the local \ac{PMIx} server library upon completion of the global collective. Data received from each node \emph{shall} be simply concatenated to form an aggregated unit, as shown in the following example:
```
gvallee: It seems that we do not emphasis *must* in the rest of the document so we should not do it here.

Chap_API_Server.tex
```diff
@@ -1300,7 +1300,21 @@ \subsection{\code{pmix_server_fencenb_fn_t}}
...
+size_t sz_blob1, sz_blob2, sz_total;
+
+sz_total = sz_blob1 + sz_blob2;
+total = (uint8_t*)malloc(sz_total);
+memcpy(total, blob1, sz_blob1);
+memcpy(\&total[sz_blob1], blob2, sz_blob2);
+\end{codepar}
+\cspecificend
+
+Note that the ordering of the data \emph{blobs} does not matter.
```
gvallee: Why do we emphasis *blobs* here? I do not see anything similar in the rest of the standard.

Chap_Introduction.tex
gvallee: I would personally suggest to leave the duplication in v3.x and remove PMIX_NUM_SLOTS in v4.

rhc54: I have restored it, clarifying that it is duplicative. I did not deprecate it as it probably just isn't worth it.

Issue #168, opened 2019-02-19
pmix_server_fencenb_fn_t and PMIX_OPERATION_SUCCEEDED


krempel-pt:
Hello once again.

This is a question on PMIx v3.1.

In the reference implementation if PMIx_Fence is called for only processes on one node, the pmix_server_fencenb_fn is called by the server library, even if it could handle the whole fence by itself.

First Question: Is this intended behaviour and if so, why? Is there a case that I do not see right now that makes it necessary to inform the host server?

Second Question: If the host server sees, that only its own node is involved, it wants to return immediately with PMIX_OPERATION_SUCCEEDED. The standard allows that in general and says in that case the callback is not to be called. I wonder now if this is valid if data was provided to be collected by the fence. Of cause the server library has its own data already so it could work that way. Perhaps a concretion in the standard at this point would be meaningful, too.

rhc54:
> First Question: Is this intended behaviour and if so, why? Is there a case that I do not see right now that makes it necessary to inform the host server?

It isn't required - just an optimization we didn't bother to implement. It would seem an unusual case that would require adding a check of both the nspace and rank values to the path, thus impacting all use-cases. Still, we could look at it some time.

> Second Question: If the host server sees, that only its own node is involved, it wants to return immediately with PMIX_OPERATION_SUCCEEDED. The standard allows that in general and says in that case the callback is not to be called. I wonder now if this is valid if data was provided to be collected by the fence. Of cause the server library has its own data already so it could work that way. Perhaps a concretion in the standard at this point would be meaningful, too.

You are correct - the data provided to the host has obviously already been collected by the PMIx server library. Thus, there is no need for it to be "returned" and the host is welcome to return "operation_succeeded". I can add some words to the standard to clarify that use-case.

PR #169, opened 2019-02-20
Clarify that fence may not involve server

rhc54:
Note that PMIx libraries may choose to implement an
optimization for the case where only the calling process
is involved in the fence operation by immediately
returning PMIX_OPERATION_SUCCEEDED from the client's
call in lieu of passing the fence operation to a PMIx server.

### DIFF
--- a/Chap_API_Key_Value_Mgmt.tex
+++ b/Chap_API_Key_Value_Mgmt.tex
@@ -612,7 +612,9 @@ The following attributes are optional for host environments that support this op
 \adviceimplstart
 We recommend that implementation of the \refattr{PMIX_TIMEOUT} attribute be left to the host environment due to race condition considerations between completion of the operation versus internal timeout in the \ac{PMIx} server library. Implementers that choose to support \refattr{PMIX_TIMEOUT} directly in the \ac{PMIx} server library must take care to resolve the race condition and should avoid passing \refattr{PMIX_TIMEOUT} to the host environment so that multiple competing timeouts are not created.
 
-Note that \ac{PMIx} libraries may choose to implement an optimization for the case where only the calling process is involved in the fence operation by immediately returning \refconst{PMIX_OPERATION_SUCCEEDED} from the client's call in lieu of passing the fence operation to a \ac{PMIx} server.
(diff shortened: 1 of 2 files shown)


rhc54:
Fixes #168

Chap_API_Key_Value_Mgmt.tex
```diff
@@ -611,6 +611,8 @@ \subsection{\code{PMIx_Fence_nb}}
 
 \adviceimplstart
 We recommend that implementation of the \refattr{PMIX_TIMEOUT} attribute be left to the host environment due to race condition considerations between completion of the operation versus internal timeout in the \ac{PMIx} server library. Implementers that choose to support \refattr{PMIX_TIMEOUT} directly in the \ac{PMIx} server library must take care to resolve the race condition and should avoid passing \refattr{PMIX_TIMEOUT} to the host environment so that multiple competing timeouts are not created.
+
+Note that \ac{PMIx} libraries may choose to implement an optimization for the case where only the calling process is involved in the fence operation by immediately returning \refconst{PMIX_OPERATION_SUCCEEDED} from the client's call in lieu of passing the fence operation to a \ac{PMIx} server.
```
jjhursey: Is this also true if all of the fence participants are on the same node? Just curious since there is no data to exchange between servers the upcall into the RM is not necessary, but the server side PMIx library needs to coordinate the local distribution of data.

rhc54: I don't believe that should be allowed as the "fence" isn't just about passing data - it is an actual synchronizing barrier. This optimization only works for the case of a single process because there is no inter-process barrier to execute.

rhc54: Do you think we need to add wording to make that clear?

jjhursey: I think it's ok as it is.

I was just thinking about the "all participants on the same node" case from the RM's perspective. In which case the synchronization (and possible data exchange) could happen all in the PMIx library without the upcall into the RM. The PMIx library might execute code on its internal server side, but there might not be a need to trigger the upcall to the RM (unless the RM wanted it for tracking purposes). Just something to think about I guess.

rhc54: Hmmm...I think maybe I do need to add a few words here. The optimization being discussed involved resolving the fence in the PMIx client library, thus avoiding any communication to the PMIx server library. This can only be done for the 1-proc case due to the barrier requirement.

If a PMIx server notes that all participants are on the local node, then it has no need to relay the fence operation to its host environment as there is no cross-node operation required. This is a separate optimization that I'll add to the doc. IIRC, we already implement that one (but not the 1-proc case).

Issue #174, opened 2019-03-28
Publish default PMIX_RANGE is up to implementation but default unpublish is SESSION


dsolt:
Maybe this is a bit picky, but it is a little odd to me that the default range for publish is up to the implementation, but the unpublish range is PMIX_RANGE_SESSION by default.  This means that a code that calls PMIx_Publish(kvs, 1) and PMIx_Unpublish(keys,NULL,0) may see the Unpublish fail if the implementations default range for PUBLISH is not PMIX_RANGE_SESSION.

dsolt:
So when I typed this in, I mistyped the args to publish, but in thinking about it, I wonder if PMIx_Publish should have been:

PMIx_Publish(info_to_publish, num_info_to_publish, info_to_influence_behavior, num_info_to_influence_behavior).   

Just a thought in case a group of people come along and decide to break backward compatibility and/or change API's dramatically. cough, cough

rhc54:
> Maybe this is a bit picky, but it is a little odd to me that the default range for publish is up to the implementation, but the unpublish range is PMIX_RANGE_SESSION by default.

I'm not seeing that in v3.1 of the standard - it states the following on page 140:
```
Publish the data in the info array for subsequent lookup. By default, the
data will be published into the PMIX_SESSION range and with
PMIX_PERSIST_APP persistence.
```
Can you point me to where you see it being left to the implementation?

> PMIx_Publish(info_to_publish, num_info_to_publish, info_to_influence_behavior, num_info_to_influence_behavior).

Yeah, that's a good point...sigh.

dsolt:
I don't know if I was looking at an old version, but I do now clearly see that Publish does say the default range is PMIX_RANGE_SESSION.   I'm not sure where I thought I saw a statement indicating that the implementation could determine this, but I re-read everything and it looks ok.   I think it is a bit confusing whether one process can publish something at the NODE level and another process does a lookup at the SESSION level or at an unspecified range and it should not find it.   I think that is true.. published and lookup data must 100% match on range specification.   I guess I'd like to see that more explicitly called out, but otherwise this actually looks good

Issue #173, opened 2019-03-27
PMIX_RANGE duplicate publish behavior unclear


dsolt:
The standard says, "Duplicate keys within the specified data range may lead to unexpected behavior depending on host RM implementation of the backing key-value store."    This implies to me that duplicate keys on different specified data ranges have a well behaved behavior.  I suggest we either change this statement to:   

"Duplicate keys may lead to unexpected behavior depending on host RM implementation of the backing key-value store." 

OR 

we need to clarify what happens if the same key is published at different levels.  Is there an implied hierarchy of ranges?   Some seem obvious, but others like PMIX_RANGE_NAMESPACE and PMIX_RANGE_LOCAL are not so obvious which one should be checked first and then PMIX_RANGE_CUSTOM of course messes everything up.

rhc54:
Hmmm...the idea was that keys published on different ranges should be orthogonal to each other, but you make a good point - we/I had overlooked the fact that the ranges overlap. I think that trying to specify our way thru the labyrinth of precedence combinations is too onerous - I think your first proposal makes the most sense to me, though it leaves uncertainty. On the other hand, that might not be bad as it might discourage people from using duplicate keys!

dsolt:
Yes, I don't see any good option here except to say that publishing keys at multiple ranges is undefined.

rhc54:
I've been trying to clarify the standard based on this ticket plus the comments on #174 and it seems to me that the two are contradicting each other. If we require that the publish and lookup ranges exactly match, then we are saying that a key must only be unique within a specified range. Effectively, we are stating that the "range" defines a key-value store. Thus, something published into the ```PMIX_RANGE_LOCAL``` store cannot be accessed by a lookup searching the ```PMIX_RANGE_SESSION``` store.

On the other hand, if we say a key must be unique across all ranges (i.e., no duplicates published to different ranges), then we are saying there is a single "global/distributed" key-value store and the lookup should return a key if the search and publishing ranges intersect. In this case, one would have to question the value of having a ```range``` specification in both the publish and lookup APIs.

I believe, therefore, we should go with the first option - a key must be unique within the specified publishing range, and the lookup range must exactly match the published range. This uniqueness guarantee is easier for users to meet and it makes the search algo a little simpler.

Make sense?

dsolt:
I've been thinking about this as well and I think that is the best way to go.   Each range is a very distinct and the publisher and lookup'er must agree on the range.

PR #176, opened 2019-04-03
Clarify relationship of range between publish and lookup

rhc54:
A key must be unique within the specified publishing range,
and the lookup range must exactly match the published range.
This uniqueness guarantee is easier for users to meet and
it makes the search algo a little simpler.


rhc54:
Refs #173

rhc54:
Refs #174 as well, though mostly in the comments

PR #180, opened 2019-04-07
Add Python Bindings Appendix

rhc54:
Create the C <-> Python datatype correspondence table.
Provide Python-language definitions for all C-level interfaces.

Fix minor typos in server module function descriptions 


rhc54:
@jjhursey See what you think of the table

App_Python.tex
```diff
@@ -0,0 +1,79 @@
...
+\chapter{Python Bindings}
+\label{app:python}
+
+While the \ac{PMIx} Standard is defined in terms of C-based \acp{API}, there is no intent to limit the use of \ac{PMIx} to that specific language. Support for other languages is captured in the Standard by describing their equivalent syntax for the \ac{PMIx} \acp{API} and native forms for the \ac{PMIx} datatypes. This Appendix specifically deals with Python interfaces, starting by a review of the \ac{PMIx} datatypes.
+
+
+%%%%%%%%%%%
+\section{Datatype Definitions}
+
+With only a few exceptions, the C-based \ac{PMIx} datatypes defined in \chapterref{chap:struct} directly translate to Python. However, Python lacks the size-specific value definitions of C (e.g., \code{uint8_t}) and thus some care must be taken to protect against overflow/underflow situations when moving between the languages. Table\ref{app:python:ctopy} lists the correspondence between datatypes in the two languages.
```
jjhursey:
```suggestion
With only a few exceptions, the C-based \ac{PMIx} datatypes defined in \chapterref{chap:struct} directly translate to Python. However, Python lacks the size-specific value definitions of C (e.g., \code{uint8_t}) and thus some care must be taken to protect against overflow/underflow situations when moving between the languages. Table~\ref{app:python:ctopy} lists the correspondence between datatypes in the two languages.
```

App_Python.tex
```diff
@@ -0,0 +1,79 @@
...
+    \code{uint, uint8_t, uint16_t, uint32_t} & PMIX_UINT, PMIX_UINT8, PMIX_UINT16, PMIX_UINT32, PMIX_UINT64 & 'uint', 'uint8', 'uint16', 'uint32', 'uint64' & Integer - if interfacing to a C-based library, care must be taken to limit the value to its corresponding range \\ \hline
+    \code{float, double} & PMIX_FLOAT, PMIX_DOUBLE & 'float', 'double' & float - if interfacing to a C-based library, care must be taken to limit the value to its corresponding range \\ \hline
+    \code{struct timeval} & PMIX_TIMEVAL & 'timeval' & Either a Python \emph{tuple} containing (sec, microsec) where each field is an integer value, \textbf{OR} a single integer value for the seconds field where the microsecond field shall be assumed to be zero \\ \hline
+    \code{time_t} & PMIX_TIME & 'time' & Must be positive integer \\ \hline
+    \refstruct{pmix_data_type_t} & PMIX_DATA_TYPE & 'dtype' & Integer - if interfacing to a C-based library, care must be taken to limit the value to a \code{uint16_t} range \\ \hline
+    \refstruct{pmix_status_t} & PMIX_STATUS & 'status' & Integer  \\ \hline
```
jjhursey: Is the Python name (in this case `status`) namespaced something like `PMIx.status`?
I suspect so, but it's not clear.

Are all of the Python names part of the PMIx namespace or are some native to Python?

rhc54: I need to add some text to better explain it as you have misunderstood what I mean by "Python name". In contrast to C, there is no reason to define new Python datatypes - in fact, it makes programming much harder if you do so. Instead, what I'm showing in the table is what you need to put into (for example) a 'value' object so we know what datatype is intended. Thus, if you have a 'value' object of ('status', 1), we know that we should interpret the integer in that object as being a pmix_status_t when converting to a C library.

jjhursey: Ok. I think I understand what you are doing now. Maybe a small example would help in the text. Something like: here is the C snippet and here is the same Python snippet with datatype X

Maybe with a `PMIx_Get` of a job level item - nothing too involved just enough to illustrate.

rhc54:
@pmrich @dsikich  Please take a look

rhc54:

          we know your current thoughts

          want to be an advocate for it
          other technologies instead

Chap_API_Server.tex
```diff
@@ -2462,7 +2462,7 @@ \subsection{\code{pmix_server_job_control_fn_t}}
                              const pmix_proc_t *requestor,
                              const pmix_proc_t targets[], size_t ntargets,
                              const pmix_info_t directives[], size_t ndirs,
-                             pmix_info_cbfunc_t cbfunc, void *cbdata)
+                             pmix_op_cbfunc_t cbfunc, void *cbdata)
```
jjhursey: Is this signature change necessary for Python or was it always supposed to be `pmix_op_cbfunc_t`?

The description in the text says it should be `pmix_op_cbfunc_t`, but the code n PMIx master the signature of these two functoins look like [this](https://github.com/pmix/pmix/blob/master/include/pmix_server.h#L348-L358):
```
typedef pmix_status_t (*pmix_server_job_control_fn_t)(const pmix_proc_t *requestor,
                                                      const pmix_proc_t targets[], size_t ntargets,
                                                      const pmix_info_t directives[], size_t ndirs,
                                                      pmix_info_cbfunc_t cbfunc, void *cbdata);
typedef pmix_status_t (*pmix_server_monitor_fn_t)(const pmix_proc_t *requestor,
                                                  const pmix_info_t *monitor, pmix_status_t error,
                                                  const pmix_info_t directives[], size_t ndirs,
                                                  pmix_info_cbfunc_t cbfunc, void *cbdata);
```
The [RFC 15](https://github.com/pmix/RFCs/blob/master/RFC0015.md) also has it as an `pmix_info_cbfunc_t` which makes me think that the description in the standard is incorrect. But I don't know if there was a deeper reason why this needed to change.

App_Python.tex
jjhursey: I didn't see an example for this in the appendix, but maybe I overlooked it. A clarifying example would be helpful.

Chap_API_Server.tex
rhc54: This is strange - I initially changed it because I thought the description said `op_cbfunc`. However, I just reset that commit out - and now I can't find any reference to the `op_cbfunc` for those functions any more. I agree that we should leave it as-is, and I guess I was just reading it all wrong. Thanks for pointing it out.

rhc54:
@jjhursey See example 😄

jjhursey:
Thanks. That example helps!

App_Python.tex
```diff
@@ -0,0 +1,3064 @@
...
+        \hline
+        C-Definition & PMIx Name & Python Definition & Notes \\ \hline
+        \endhead
+        \code{bool} & PMIX_BOOL & boolean & \\ \hline
+        \code{byte} & PMIX_BYTE & A single element byte array (i.e., a byte array of length one) & \\ \hline
+        \code{char*} & PMIX_STRING & string & \\ \hline
+        \code{size_t} & PMIX_SIZE & integer & \\ \hline
+        \code{pid_t} & PMIX_PID & integer & value shall be limited to the \code{uint32_t} range \\ \hline
+        \code{int, int8_t, int16_t, int32_t, int64_t} & PMIX_INT, PMIX_INT8, PMIX_INT16, PMIX_INT32, PMIX_INT64 & integer & value shall be limited to its corresponding range \\ \hline
+        \code{uint, uint8_t, uint16_t, uint32_t} & PMIX_UINT, PMIX_UINT8, PMIX_UINT16, PMIX_UINT32, PMIX_UINT64 & integer & value shall be limited to its corresponding range \\ \hline
```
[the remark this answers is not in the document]
rhc54: Got it - thx!

PR #193, opened 2019-06-26
Proposed PMIx Standards Process

rhc54:


README.md
```diff
@@ -2,24 +2,198 @@
...
+</div>
+\*  Co-Chair (even year) &nbsp;&nbsp;&nbsp;&nbsp; \** Co-Chair (odd year)
+
+### General voting rules
+
+The philosophy and goal of the ASC is to reach consensus on all, if not most, issues. In the event that full consensus cannot be reached, the ASC will use the following rules:
+
+* Each active ASC Member organization will receive one vote. All votes are required to pass with 2/3 majority. A quorum (>50%) of Members must be present (either physically or virtually). Abstentions will not be counted in the total votes when calculating whether there is a 2/3 majority.
+
+* A Member may submit a vote _in absentia_ to the ASC Co-Chairs in advance of the meeting if attendance by a representative is not possible (e.g., a quarterly meeting that falls during a national holiday period), but each ASC Member organization can only cast one vote. This prevents a Member from voting on behalf of its own entity and acting as a proxy for another member.
```
SteVwonder: The latter part of this is a little confusing to me.  Who ends up voting on behalf of the absent member?  The co-chair? A second representative from a different member institution?

README.md
```diff
@@ -2,24 +2,198 @@
...
+
+    All proposed additions to the PMIx Standard enter the PMIx Standard under Provisional status. Due to the flexible nature of the Provisional class, no guarantees are made with respect to backward or cross-version compatibility for entries in this class.
+
+* "**Stable**" entries are permanent, non-changing parts of the PMIx Standard subject solely to deprecation. In this class, full guarantees are provided with respect to backward and cross-version compatibility beginning with the version upon which the entry was granted "stable" status. No guarantees are made with respect to versions that included the entry during its "provisional" status. Note that stable APIs may continue to define new supported key-value attribute pairs but may not remove any previous definitions. Thus, the _behavior_ of a stable API may be _extended_ via a new attribute but an existing behavior directed by a current attribute may not be eliminated or modified.
```
SteVwonder: Just to clarify, does this mean you can only deprecate whole interfaces and cannot deprecate an individual attribute?

README.md
```diff
@@ -2,24 +2,198 @@
...
+
+#### Proposed modification
+When a specific change to the PMIx Standard has been identified a PR must be opened with a reference made to the corresponding Issue(s). When the PR is opened it should be placed in a [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) state until the following conditions are met:
+
+* The contents of the PR are ready for review and broad discussion with the PMIx community. Note that discussion over the PR is encouraged at all times even while in draft form.
+
+* An open-source implementation of the proposal is available and a reference to the implementation included in the PR. Note that a complete implementation is not necessarily required to meet this condition - e.g., pseudocode detailing the execution path supporting the proposed modification can be provided.More complete prototype implementations may be requested by the PMIx community when proposals involving significant changes in behavior are made.
```
SteVwonder: Is the idea of allowing psuedocode to better support closed-source implementations that do not want to release their full implementation of a feature and instead provide a rough sketch?

Minor nit: missing a space after a full stop
```suggestion
* An open-source implementation of the proposal is available and a reference to the implementation included in the PR. Note that a complete implementation is not necessarily required to meet this condition - e.g., pseudocode detailing the execution path supporting the proposed modification can be provided. More complete prototype implementations may be requested by the PMIx community when proposals involving significant changes in behavior are made.
```

tonycurtis:
@rhc54 very nice.

README.md
rhc54: I'll clarify it. The short answer is: nobody votes on their behalf - their vote is simply recorded as if they were there and cast it themselves, with a notation that the vote was cast "in absentia"

README.md
rhc54: The pseudocode idea was in fact driven by the closed-source concern (not just an entire implementation - they might have a proprietary plugin that implements the change), but it also seemed to me that we don't always require an in-depth coding example right away (e.g., for some fairly trivial modification). So someone could provide us initially with pseudocode and the Members can decide if a more detailed implementation needs to be seen prior to provisional approval.

I don't have strong feelings on it - any thoughts?

README.md
rhc54: Hmmm...good point. It does get a little complicated when considering the cross-version issue for a stable definition.

I guess I would propose that we allow deprecation of an individual attribute, but that the deprecation proposal must also include deprecation of the associated behavior for every impacted API. Thus, you don't only deprecate an attribute - you are deprecating all behaviors associated with that attribute.

Make sense?

README.md
rhc54: I clarified this:

> A Member may submit a vote _in absentia_ to the ASC Co-Chairs in advance of the meeting if attendance by a representative is not possible (e.g., a quarterly meeting that falls during a national holiday period), but each ASC Member organization can only cast one vote. This prevents a Member from voting on behalf of its own entity and acting as a proxy for another member. _In absentia_ votes are counted by the Co-Chairs as if the member were present (thus counting towards the meeting's quorum requirement), but are recorded as having been voted _in absentia_ in the official tally.

README.md
rhc54: I tried to clarify this - see what you think:

In the class definitions:

> * "**Stable**" entries are permanent, non-changing parts of the PMIx Standard subject solely to deprecation. In this class, full guarantees are provided with respect to backward and cross-version compatibility beginning with the version upon which the entry was granted "stable" status. No guarantees are made with respect to versions that included the entry during its "provisional" status. Note that Stable APIs may continue to define new supported key-value attribute pairs, thereby extending the _behavior_ of a stable API without modifying its signature.

>    Attributes used by APIs in the "Stable" class must also reside in that class. A "Stable" API that wishes to utilize a "Provisional" attribute must therefore first seek elevation of the attribute to the "Stable" class. This is necessary to ensure that all "Stable" parts of the Standard are capable of meeting the compatibility requirements. Provisional" APIs are, of course, free to utilize any "Stable" attribute in addition to anything in the "Provisional" class.

in the elevation to stable status section:

> Note that attributes can be proposed for elevation separate from the APIs that use them. Promotion of one or more attributes independent of any API may, for example, occur when a "Stable" API has expressed a desire to utilize them. In such cases, the proposal should clearly identify the "Stable" APIs making the request and provide some justification for it.

and down in the deprecation section:

> Deprecation of "Stable" attributes can be proposed separately from the APIs that support them - i.e., it is not required that an API which utilizes an attribute be deprecated at the same time as the attribute. The process of deprecating an attribute is, however, somewhat more complicated as it may (a) be used in multiple "Stable" as well as "Provisional" APIs and (b) impact the behavior of an API that supports it. Care must be taken, therefore, to identify _all_ places in the Standard that are impacted by the deprecation proposal. Where "Stable" APIs are impacted, a deprecation warning for the change in _behavior_ must be included along with warning of deprecation of the attribute itself.

SteVwonder: Thanks @rhc54. Yeah. I think I understand, but just to make sure I'm not misunderstanding the terminology, I'd like to give a concrete example and see if it aligns with what you mean.  If `PMIx_Process_monitor` is marked as "Stable", then all of the attributes (optional or required) that can be passed into `PMIx_Process_monitor` (like `PMIX_MONITOR_HEARTBEAT` and `PMIX_MONITOR_FILE`) must also be "Stable", correct?

What about the scenario where someone wants to extend `PMIx_Process_monitor` to include a new monitoring type besides heartbeat or file, like maybe `PMIX_MONITOR_NETWORK` for example.  Would this new attribute have to be immediately added to PMIx as "Stable" to be usable with `PMIx_Process_monitor`?  Or can it start as "Provisional" and then be moved to "Stable"?

rhc54: You ask a good question that highlights the fact that I left out the attribute "required" vs "optional" explanation. Here is what I added to the "classes" section to help clarify:

At the beginning of that section:

> The PMIx Standard is composed of a combination of APIs and key-value attributes. APIs are intentionally designed to be "lightweight" - i.e., the API signature is generic with a minimal number of fixed parameters in its signature. With few exceptions, APIs are required to include an array of key-value attributes in their signature to facilitate behavioral extensions without modifying the API signature. Thus, PMIx attempts to minimize deprecation and modification of APIs as use models evolve over time.

> Each API can designate a given attribute as either "**required**" or "**optional**" - this indicates whether or not a PMIx implementation of that API is required to support the attribute. PMIx implementations are requested to support optional attributes, but are not required to do so. This is intended to both provide flexibility to PMIx implementors (e.g., to tailor the implementation to their target audience) and to ease adoption of the Standard.

> Attributes can similarly be designated by the caller of a PMIx API as either "required" or "optional". This serves as a directive to the PMIx implementation as to the expected response should the implementation not support that attribute. Attributes provided by the caller that are marked as "required" but are not supported by the API must result in the API returning "not supported". Provided attributes marked as "optional" may be ignored if not supported.

Under "provisional":

> All proposed additions to the PMIx Standard enter the PMIx Standard under Provisional status. Attributes in the "Provisional" class may only be designated as "optional" by an API - i.e., an API definition cannot designate a "Provisional" attribute as "required". Users, however, are always free to "require" that an attribute be supported when calling the API, thus directing the API's response should the attribute not be supported by it.

Under "stable":

> Stable APIs may continue to define and/or adopt new supported key-value attribute pairs, thereby extending the _behavior_ of a stable API without modifying its signature. As previously specified, attributes in the "Provisional" class can only be specified as "optional" - therefore, _required_ attributes used by APIs in the "Stable" class must also reside in that class. A "Stable" API that wishes to utilize a "Provisional" attribute must either first seek elevation of the attribute to the "Stable" class so it can be "required", or must designate that attribute as "optional". This is necessary to ensure that all "Stable" parts of the Standard are capable of meeting the compatibility requirements. "Provisional" APIs are, of course, free to utilize any "Stable" attribute in addition to anything in the "Provisional" class.

SteVwonder: Ah! That makes a lot of sense.  Thanks @rhc54!

README.md
SteVwonder: If I were reviewing an extensions for inclusion under the "Provisional" class, I believe that I would be OK with psuedo-code.  For moving to the "Stable" class, I would feel very strongly about requiring an open-source implementation be referenced, so that I could take the implementation into account (how difficult it appears to correctly implement, does the documented semantics align with the implemented semantics, etc).  I'm not sure if others agree enough that we want to document that as a hard requirement, but that is at least my thoughts.

rhc54: Yeah, I agree. However, I'm not sure how to handle proprietary implementations in that scenario. Probably something we should discuss in tomorrow's WG meeting.

README.md
```diff
@@ -2,24 +2,214 @@
...
+* Plan release timelines for versions of the PMIx Standard based on available resources, target features, and target goals of the Member organizations.
+* Vote on all issues requiring community consensus (e.g., acceptance of proposed modifications to the Standard), one vote per Member organization. In addition to voting on proposed modifications, the ASC can be requested to vote on any issue submitted by a Member for consideration.
+* Assemble working groups to address specific areas of interest to the community.
+* Nominate and approve the appointment of individuals to fill the role of Release Manager (per release), as described below.
 
-If you have a question about the PMIx interface or want to raise a question for discussion then please file an Issue against this repository. We are using the Issues to discuss topics, and will link Pull Requests related to that topic to that Issue. This provides the community with a public, archived forum for discussion.
+The ASC is led by two Co-Chairs responsible for organizing and conducting the quarterly meetings. This includes:
 
-### Participatin in the writing
+* Announcing meeting agendas at least one month in advance of the meeting
```
jjhursey: To this point, we probably need to capture the notion that PRs targeted for consideration at the meeting should be in a 'ready' state at 1 month out so that they can be identified and added to the agenda. That gives folks a month to review the PRs and prepare for the meeting.

Maybe something like this (it's rough)
> ASC Co-Chairs must send out an agenda 1 month before the meeting. Any PRs that wish to be considered for that meeting [must | should] be in the "Reading" (for Provisional class changes) or "Eligible" (for Stable class and Deprecation changes) state at this time to make it on the agenda. The ASC Co-Chairs may allow for PRs to be added to the agenda after this time on a case-by-case basis.

README.md
```diff
@@ -2,24 +2,214 @@
...
+
+
+#### Deprecation
+Deprecation of one or more existing elements of the Standard, be they APIs or attributes, starts by mirroring the path of a newly proposed modification. An Issue proposing the deprecation and explaining the reasons behind it shall be opened for initial discussion. If desired, this may immediately be accompanied by the associated PR to remove the identified element(s), or the PR can be generated after an appropriate amount of discussion has concluded. The deprecation PR must include the required "**DEPRECATION**" warning language to be added to the Standard - examples are provided in the GitHub Deprecation Template. The same "**Straw Poll**" voting mechanism, including presentation at a quarterly meeting, shall be used to determine if the proposal merits deeper consideration by the ASC.
```
jjhursey: > "this may immediately be accompanied by the associated PR to remove the identified element(s)"

I think we first need a PR to mark the API as deprecated that the RMs can commit right away. Then a second PR which will remove it. This allows us to release a standard A.0 with the API labeled as deprecated but still included in the document. Then at C.0 it can be formally removed.

Maybe the timeline can be:
 * A.0 : API/Attribute marked as deprecated in place in the document. Update change listing to indicate it being added as newly deprecated. PR specific to this change by the authors.
 * B.0 : API/Attribute marked as deprecated in place in the document. Update change listing to indicate possible removal in next major version. PR specific to this change by the RMs (since it's a changelog item).
 * C.0 : API/Attribute removed from the document. Update change listing to indicate removal from the standard. PR specific to this change by the authors & RMs.

README.md
```diff
@@ -2,24 +2,214 @@
...
+### Release Managers
+
+One or more Release Managers are assigned per release series of the PMIx Standard. Coordination and scheduling of releases in the series is delegated to the Release Manager(s) by the ASC, subject to guidance regarding needs on the part of Members. Release Managers have the following duties:
+
+* For each release, lead the establishment of release criteria by the ASC. This can include a list of proposed modifications to be decided upon prior to release and target release dates
+* Shepherd progress towards completing the schedule through interactions with proposal authors
+* Maintain the Revisions section of the document to ensure accurate communication of what has changed in each release
+
+Note that Release Managers are not precluded from proposing modifications to the Standard (including to the series under their management) during the course of their assignment. In addition, Release Managers are not responsible for promoting acceptance of any proposal nor contributing to the completion of a proposal in order to meet schedule.
+
```
jjhursey: Should we grant some liberty to the ASC Co-Chairs and Release Managers to work through any merge conflicts that might arise after a quarterly meeting?

As long as there are no significant issues in the merge they should be able to resolve this without more process - using their discretion. If there is a significant issue then they can bring it to the community with a suggested resolution.

jjhursey:
I tried to organize a timeline sketch of the process below. Is this an accurate summary of the process?

Provisional Status timeline (Shortest path to acceptance - 1 quarterly meeting)
 (1) Open a Draft PR while writing
 (2) Remove Draft PR status when ready for discussion
 (3) Discuss PR - only add new commits
 (4) Label "RFC" and announce to mailing list it's ready for formal review
 (5) Add a "Straw Poll to move to Reading" comment
 (6) ASC Co-Chairs schedule proposal for "Reading" at next quarterly meeting
 (7) Proposal read at quarterly meeting
 (8) Add a "Straw Poll to move to Acceptance" comment
 (9) After 2 weeks, based on Straw Poll the PR is either "Accepted" or "Pushed Back"
 (10) Accepted proposals merged by Release Managers

Stable Status timeline (Shortest path to acceptance - 2 quarterly meetings)
 (1) Open a PR with justification for status change
 (2) Add a "Straw Poll" comment
 (3) If passes Straw Poll then ASC Co-Chairs label it as "Eligible" in next quarterly meeting
 (4) Author presents Eligible PR in quarterly meeting
 (5) Formal vote taken to designate as "Pushed Back" or "First Pass"
 (6) If passes, then author presents it again in the following quarterly meeting.
 (7) Formal vote taken to designate as "Pushed Back" or "Accepted as Stable"
 (8) Accepted proposals merged by Release Managers

Deprecation timeline (Shortest path to acceptance - 1 quarterly meeting)
 (1) Open an Issue for discussion
 (2) PR filed to mark API as deprecated
 (2) Add a "Straw Poll" comment
 (3) If passes Straw Poll then ASC Co-Chairs label it as "Eligible" in next quarterly meeting
 (4) Author presents Eligible PR in quarterly meeting
 (5) Formal vote taken to designate as "Pushed Back" or "Deprecated"
 (6) Deprecated proposals merged by Release Managers

README.md
rhc54: I guess I thought that was communicated elsewhere where we talk about submitting a PR for consideration, but I can add a label to make it clearer.

README.md
rhc54: Not sure I understand your comment - the paragraph states that a PR is required. It can either be generated immediately upon opening the Issue, or after some period of discussion. Is that not clear?

Or is your comment not so much about the PR but more about wanting two separate PRs - one to deprecate and one to remove? I guess I can see that so the PR can contain the actual change.

README.md
rhc54: 😆 I guess I continue to mistakenly assume at least a minimal level of common sense is being applied. Sure, we can spell it out. Just remember: the more you try to over-think the process and make everything "rigid", the more corner-cases you create that challenge anyone's ability to apply common sense.

rhc54:
> Provisional Status timeline (Shortest path to acceptance - 1 quarterly meeting)
(1) Open a Draft PR while writing
(2) Remove Draft PR status when ready for discussion
(3) Discuss PR - only add new commits
(4) Label "RFC" and announce to mailing list it's ready for formal review
(5) Add a "Straw Poll to move to Reading" comment
(6) ASC Co-Chairs schedule proposal for "Reading" at next quarterly meeting
(7) Proposal read at quarterly meeting
(8) Add a "Straw Poll to move to Acceptance" comment
(9) After 2 weeks, based on Straw Poll the PR is either "Accepted" or "Pushed Back"
(10) Accepted proposals merged by Release Managers

Not exactly - I see no reason for _two_ straw polls.  The author(s) should simply put it up for review at the next quarterly meeting whereupon a straw poll commences. After the presentation at the meeting, members can either revise their prior poll votes or, if they haven't already voted, can cast their vote. No reason to make people re-vote if their opinion didn't change.

README.md
rhc54: Here is what I added:

> * Announcing meeting agendas at least one month in advance of the meeting. Any proposal wishing to be considered for the next meeting must be put forward (by assigning the proper label, as described below) prior to that time. The ASC Co-Chairs may allow a proposal to be added to the agenda after this time on a case-by-case basis at their sole discretion.

README.md
rhc54: Here is what I modified in the deprecation section:

> If desired, this may immediately be accompanied by the associated PR to mark the identified element(s) as deprecated, or the PR can be generated after an appropriate amount of discussion has concluded.

...

> At the conclusion of the deprecation cycle, a second "removal" PR shall be submitted to actually remove the deprecated element(s) from the PMIx Standard document. When ready for consideration, the author(s) shall label the PR with the "Removal" tag, thereby requesting that the matter be placed on the agenda of the next quarterly meeting (subject to the one month deadline). The "straw poll" approval process utilized for "Provisional" proposals is utilized to ensure adequate review is performed to avoid inadvertent removal of elements not included in the original deprecation proposal. Once the process has been concluded, the Release Manager shall commit the change in accordance with their schedule.

README.md
rhc54: Here is what I changed it to:

> Note that Release Managers are not precluded from proposing modifications to the Standard (including to the series under their management) during the course of their assignment. In addition, Release Managers are not responsible for promoting acceptance of any proposal nor contributing to the completion of a proposal in order to meet schedule, and may resolve any merge conflicts resulting from commit of approved changes without further process.

README.md
jjhursey: Looks good to me.

README.md
jjhursey: Yeah, 1 PR to depreciate, and 1 PR to remove. The wording that was there was that you file 1 PR to remove meaning that the RMs have to figure out all of the places in the document to place a deprecation warning. The depreciation PR also forces the authors to identify all of the places where that API/Attribute is used and add some warning about depreciation in the future. This can help the reviewers understand the full extent of the deprecations impact.

What you have is good. I would suggest changing the first block to make it a bit more clear that a PR is required for deprecation before voting. The wording you have makes the PR optional. Maybe something like:
> The authors must file a Depreciation PR for voting which identifies the specific changes to the document where deprecation warnings must be added and/or wording needs to be modified. Discussion may occur on the Issue before a PR is generated, but a PR is necessary for voting.

jjhursey:
Where I got confused, maybe, is that the first straw poll (5) is open to everyone whereas the second iteration (8) is worded such that only ASC members can adjust their votes. That indicated in my mind that they are two different things. I'm fine (and prefer) with (8) being open to anyone since it's not an official voting mechanism. Here is the sentence:
> At the end of the agenda item, those members wishing to adjust their straw votes will be given an opportunity to do so. Members who did not attend the meeting will also have a chance to adjust their votes based on the published minutes of the meeting.

So that timeline looks like this then - modified (5) and (8)?

Provisional Status timeline (Shortest path to acceptance - 1 quarterly meeting)
(1) Open a Draft PR while writing
(2) Remove Draft PR status when ready for discussion
(3) Discuss PR - only add new commits
(4) Label "RFC" and announce to mailing list it's ready for formal review
(5) Add a "Straw Poll" comment to move the PR to Reading
(6) ASC Co-Chairs schedule proposal for "Reading" at next quarterly meeting
(7) Proposal read at quarterly meeting
(8) ASC Members may adjust their straw votes to move to the PR to Acceptance
(9) After 2 weeks, based on Straw Poll the PR is either "Accepted" or "Pushed Back"
(10) Accepted proposals merged by Release Managers

Note that I'm not advocating for more language here, or even a change is what is here (I think it's fine). I am just finding this difficult to read in such a way that I know what the "next step" in the process is (e.g., Do I send an email at this stage? What labels do I need to apply now? Am I responsible for creating the Straw Poll comment or is someone else?). So I tried to create a timeline for clarity at least in my mind.

There are a number of places in here that add justification for the process mixed with the process itself. Maybe we can separate those two so that the timeline and expectations are clearer? Or have a high-level bulleted list in each section for guidance before the body of text. I'm not sure I have a firm suggest yet.

rhc54:
Okay, I gave it a try to clarify who does what when:

> The process for acceptance to Provisional status begins when the author(s) assign an "**RFC**" label to the GitHub PR, add an empty "**Straw Poll**" comment to it, and send an announcement of the PR to the PMIx community mailing list requesting consideration of the PR for Provisional status. The announcement shall include a link to the PR, a brief description of the proposed change, and indicate any desired timeline for approval. It is especially important that proposals tied to contractual milestones be so designated to ensure that members are alerted to potential time pressures.

> The "Straw Poll" comment is used by the ASC as a mechanism for obtaining a sense of how many people have viewed the PR and are following the discussion. This serves no official purpose other than to help define participation in the consensus building process and record the following tally:

I would recommend leaving the justification with the process - otherwise, we are bound to get a constant barrage of "but why" questions as nobody will look elsewhere for the justification.

Note that I am unaware of any mechanism by which we could restrict the "straw poll" to only ASC members (at least, as currently constructed). Thus, the polls always have to be considered informal. However, the Co-Chairs can easily distinguish who cast each vote if that ever becomes a major issue, or could execute a "straw poll" via email if someone requests a secret ballot.

I have no objection to including a timeline in the doc if that helps delineate what needs to be done and when.

mike0042:
Suggestion: Rather than the second weak of the month for quarterly meetings, the third week may be better to avoid conflicts with holidays (e.g. Easter).
Suggestion: Provide a nomination process that allows members to nominate other strong contributors as members. The members then vote on adding the individual.

README.md
```diff
@@ -2,24 +2,215 @@
...
+* A Member is removed from the ASC (a) upon having failed to have at least one representative attend two quarterly meetings in the prior twelve months, or (b) by vote of the other ASC members per the voting rules defined below. Attendance by a non-voting member of the organization (i.e., someone not officially designated as an alternate for voting purposes) counts towards meeting this requirement.
+
+The ASC meets quarterly, with at least one quarterly meeting each year to be held "in-person" at an agreed-upon location unless the ASC Members vote otherwise. A vote to "virtualize" the in-person meeting can only apply to that given year. Other quarterly meetings may be held by teleconference in accordance with the desires of ASC members. ASC quarterly meetings are conducted according to the following schedule:
+
+* 2nd week of January
+* 2nd week of April
+* 2nd week of July
+* 2nd week of October
+
+ASC Members may, if desired, vote to alter the timing of the meetings in a given year to accommodate Member schedules, subject to the voting rules described below.
```
naughtont3: May want to add a bit of text to clarify when meeting schedule changes should take place.  For example during telecon, the example given was to vote to make the scheduled meeting change at or before the quarterly prior to the intended meeting reschedule.  I think we could do it during the quarter but maybe minimum of 2 months.

SteVwonder:
Note from the call today: There should be a line added about how the release cycle is tied to the quarterly meeting cycle (at most there is 1 minor release per quarterly meeting).  So an interface that is going to change after two minor releases will have at least 6 months of runway before being changed/removed.

gvallee:
Question: Assuming an API or an attribute is already provisional. Now let's assume that a modification is accepted to that provisional API or attribute, I would assume it resets the timer to be eligible for becoming stable. Correct? So every time a modification in the context of a provisional functionality is made, it should the eligibility by a **minimum** of two quarterly meetings. Is too slow? The right speed? The concern here is in the context of refining the definition and semantics of a provisional functionality: it may take a few iterations to "get the text right" and that may potentially drastically slow down the process if everything it resets the timer for stable eligibility, even if these modifications remain minor. I am not arguing that this is necessarily a bad thing (we want good text to become stable) but is it our intent and if so should that be explicitly spelled out in the document?

jjhursey:
Per Teleconf Friday, June 28, 2019
 * I captured a few notes from the discussion in the meeting minutes [here](https://github.com/pmix/pmix-standard/wiki/Meeting-2019-06-28) - most suggestions have already been posted here (please continue to do so).

rhc54:
> Note from the call today: There should be a line added about how the release cycle is tied to the quarterly meeting cycle (at most there is 1 minor release per quarterly meeting). So an interface that is going to change after two minor releases will have at least 6 months of runway before being changed/removed.

Made the following modification to the Release Manager section:

> The Release Managers shall notify the Co-Chairs when a given release is complete and ready for publication. The Co-Chairs will add an item to the agenda of the next quarterly meeting (subject to the one month deadline) for final approval to publish the document. Once the ASC has voted its approval, the Release Managers shall post the new release on the community's web site and send an email announcing the release to the community mailing list.

README.md
rhc54: Made following change:

> ASC Members may, if desired, vote to alter the timing of the meetings in a given year to accommodate Member schedules, subject to the voting rules described below. Changes to a given meeting's dates must be determined no later than the prior quarterly meeting.

rhc54:
> Suggestion: Rather than the second weak of the month for quarterly meetings, the third week may be better to avoid conflicts with holidays (e.g. Easter).

Done - moved to 3rd week of month

> Suggestion: Provide a nomination process that allows members to nominate other strong contributors as members. The members then vote on adding the individual.

Made following change:

> All new Members must either request membership or be nominated by an existing ASC Member, and be approved by the ASC according to the voting rules described below.

rhc54:
> Suggestion: Add sentence like “informal members are encouraged to attend quarterly meetings” to make it clear that the meetings are not exclusive to members.

Modified as below:

> Informal participants do not have voting privileges on administrative matters, but are encouraged to attend the quarterly meetings and participate in those discussions.

rhc54:
> > Note from the call today: There should be a line added about how the release cycle is tied to the quarterly meeting cycle (at most there is 1 minor release per quarterly meeting). So an interface that is going to change after two minor releases will have at least 6 months of runway before being changed/removed.
> 
> Made the following modification to the Release Manager section:
> 
> > The Release Managers shall notify the Co-Chairs when a given release is complete and ready for publication. The Co-Chairs will add an item to the agenda of the next quarterly meeting (subject to the one month deadline) for final approval to publish the document. Once the ASC has voted its approval, the Release Managers shall post the new release on the community's web site and send an email announcing the release to the community mailing list.

Actually, I have revised this a bit more to provide some requested emphasis on the annual f2f meeting as a means of promoting attendance:

> The Release Managers shall notify the Co-Chairs when a minor release candidate is complete and ready for publication. The Co-Chairs will add an item to the agenda of the next quarterly meeting (subject to the one month deadline) for final approval to publish the document. Once the ASC has voted its approval, the Release Managers shall post the new release on the community's web site and send an email announcing the release to the community mailing list. Minor releases, therefore, can occur on an at most a quarterly basis.

> Major release candidates must be approved during the ASC's annual face-to-face meeting. Release Managers must notify the Co-Chairs of a major release candidate a minimum of two quarters before the target approval meeting. The Co-Chairs will add an item to the agenda of each quarterly meeting for discussion and review of the document, and on the face-to-face meeting for vote on final approval. Once the ASC has voted its approval, the Release Managers shall post the new release on the community's web site and send an email announcing the release to the community mailing list. Should the ASC not approve the release, subsequent release candidates can be considered at upcoming regular ASC meetings - i.e., the major release candidate does not have to wait for the next annual face-to-face meeting for reconsideration. However, no subsequent major release may begin the approval process until after the current candidate has been published. Major releases, therefore, can occur no more than once per year.

README.md
```diff
@@ -2,24 +2,220 @@
...
+
+
+#### Elevation to Stable status
+Provisional elements of the Standard that have been included in at least two minor releases become eligible for elevation to "**Stable**" status. The elevation process begins with a PR requesting consideration of the change, identifying the specific elements (APIs and attributes) to be included, and a justification for Stable status. The PR is opened in a [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) state until ready for submission for review. Justification should be commensurate with the magnitude of the request - e.g., elevating an API to "Stable" likely requires more justification than a key-value attribute. While there is no clear formula for justifying elevation, typical considerations might include the breadth of adoption (by implementers, system management vendors, and/or programming libraries) and stability of the elements themselves (i.e., time since last modification).
```
artpol84: > The elevation process begins with a PR requesting consideration of the change, identifying the specific elements (APIs and attributes) to be included, and a justification for Stable status.

IMO an Issue seems to be the right place for this kind of discussion and accompanying PR should be fairly trivial and just change the labels in the document.

I think that both modifications and elevations should go the same route.
This way Issue tab will always be an entry point for a problem and will contain all required links inside.

README.md
```diff
@@ -2,24 +2,220 @@
...
+
+At the conclusion of the second meeting where the proposal is up for consideration, another formal vote of the ASC shall be held (per the above voting rules) to determine final disposition of the proposal. The Co-Chairs shall subsequently set a GitHub label indicating the result as either "**Stable**" (indicating that the proposed elements have been elevated to that class) or "**Pushed Back**".
+
+Proposals that are pushed back at any point can be modified and resubmitted by the author(s). However, the updated PR must pass through the complete elevation process from the beginning to ensure full consideration of the revised request. A new PR for this purpose _must not_ be submitted for the identical elements in order to retain full information regarding prior concerns and suggestions.
```
artpol84: I think that this part is missing what the release manager should do once elevation is accepted.

README.md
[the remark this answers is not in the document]
SteVwonder: I do not think so.  At least, if we follow semantic versioning, a major version change is only required when a breaking API change is made.  If it is required, then it may be years between when an interface goes from "Provisional" to "Stable".  That seems unnecessarily long to me

I do think that it requires a minor version change.

README.md
```diff
@@ -2,24 +2,220 @@
...
+
+Discussion of the PR will continue until all dissenting opinions and questions have been addressed, upon which the PR may be put forward for Provisional acceptance. Note that addressing dissenting opinions does not necessarily mean achieving complete resolution on them. A dissent is not considered to have "veto" power over a proposal, but must be addressed to a degree that convinces the majority of ASC members.
+
+#### Provisional acceptance
+The process for acceptance to Provisional status begins when the author(s) assign an "**RFC**" label to the GitHub PR, add an empty "**Straw Poll**" comment to it, and send an announcement of the PR to the PMIx community mailing list requesting consideration of the PR for Provisional status. The announcement shall include a link to the PR, a brief description of the proposed change, and indicate any desired timeline for approval. It is especially important that proposals tied to contractual milestones be so designated to ensure that members are alerted to potential time pressures.
```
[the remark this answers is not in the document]
SteVwonder: I do not think so.  My understanding is that the whole purpose of "Provisional" interfaces/attributes is to allow for breaking changes in between major versions (i.e., at minor versions).  I think it should require a minor version change though.

README.md
```diff
@@ -2,24 +2,220 @@
...
+
+* The contents of the PR are ready for review and broad discussion with the PMIx community. Note that discussion over the PR is encouraged at all times even while in draft form.
+
+* An open-source implementation of the proposal is available and a reference to the implementation included in the PR. Note that a complete implementation is not necessarily required to meet this condition - e.g., pseudocode detailing the execution path supporting the proposed modification can be provided. More complete prototype implementations may be requested by the PMIx community when proposals involving significant changes in behavior are made.
+
+Only new commits are permitted on the branch once the [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) designation has been removed.
+The branch associated with the PR must not be squashed or force pushed after this point. This preserves the timeline of changes in relationship to the discussion that may occur on the PR.
```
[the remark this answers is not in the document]
SteVwonder: > Also, it is not clear what happens if this rule is violated.

That's a good point.  I think, if anyone notices, they should call it out in the PR and inform the PR author of this rule.  Assuming the squash/force-push was done accidentally/without knowledge of this rule, I would leave the PR open.  If someone was found to have maliciously tried to sneak in a change (which I highly doubt will ever actually happen), then I think the PR should be closed.

> Should there be an option of creating a cleaner PR as the replacement for the one that has heavy discussions?

:+1: Seems reasonable to me.

README.md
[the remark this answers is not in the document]
artpol84: Though each elevation can be reviewed independently with respect to a major version bump, it feels like going from Provisional to Stable changes the strength of the API/key-value which breaks backward compatibility.
So I thought there might be a general rule about that.

README.md
```diff
@@ -2,24 +2,220 @@
...
+
+* A Member is removed from the ASC (a) upon having failed to have at least one representative attend two quarterly meetings in the prior twelve months, or (b) by vote of the other ASC members per the voting rules defined below. Attendance by a non-voting member of the organization (i.e., someone not officially designated as an alternate for voting purposes) counts towards meeting this requirement.
```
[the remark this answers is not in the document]
rhc54: >  Does that mean that 2/3 of members may kick another member if they want to?

Yes! Standards organizations typically protect themselves with a clause such as this - e.g., for the case where someone or some organization is deliberately using the standard as a competitive "weapon" or otherwise behaving in an unprofessional/disruptive manner.

README.md
rhc54: Added

README.md
rhc54: Okay, done - but keep in mind that the PR is not necessarily as trivial as you make it sound. It could require reorganizing APIs to a different "stable" section of the document, modifying attributes in multiple places to become "required", etc. It is unlikely to just involve changing a few labels.

README.md
```diff
@@ -2,24 +2,220 @@
...
+
+Anyone indicating concern must provide a comment on the ticket following the "**Straw Poll**" comment describing their concern. Sufficient detail should be provided to allow the author(s) to respond to the specific concerns. The author(s) can then work to address the concern, modifying the PR as required - modification should be accompanied by an appropriate comment citing the original participant who raised the concern to ensure they are notified of the change. Participants shall indicate their acceptance or rejection of the modification by so marking the change comment and adjusting their original "vote" on the straw poll if appropriate.
+
+The ASC Co-Chairs shall check the status of discussion on the PR on a weekly basis. Once discussion appears to have consolidated (either measured by participation in the straw poll, rate of comments, or a combination of both), the Co-Chairs will schedule the proposal for a reading at the next quarterly meeting of the ASC. The proposal shall be included in the announced agenda for the meeting along with a link to the PR and the results of the straw poll.
```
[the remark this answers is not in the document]
rhc54: Modified:

> The ASC Co-Chairs shall monitor the status of discussion on the PR.

README.md
```diff
@@ -2,24 +2,251 @@
...
 
-If you want plan to work on a large section of the document then please coordinate with the community on the PMIx mailing list to make sure we are not duplicating efforts. When working on a large section (particularly if the changes are going to take a day or two to make) then let the community know, file a PR with your first set of changes, and mark the PR as "Work In Progress".  Once it is ready for review then remove the "Work In Progress" label, and let the community know that it's ready for discussion.
+Participation in the consortium is both formal and informal:
 
-## Building the PMIx Standard document
+* Informal participation is encouraged by working with the PMIx Standard community and/or one of the PMIx implementation libraries in an _ad hoc_ manner, such as (but not limited to) engaging in discussions on the mailing lists, reporting testing results, providing feature requests and bug reports, filing pull requests for proposed modifications, etc. Informal participants do not have voting privileges on administrative matters, but are encouraged to attend the quarterly meetings and participate in those discussions.
```
dsolt: Should make a note that informal participants can attend meetings and, in fact, must attend meetings to become formal members.   Maybe something such as: "Informal participants may attend quarterly meetings and an individual or organization must attend meetings as informal participants before being elevated to Members."

README.md
```diff
@@ -2,24 +2,251 @@
...
+
+* A Member is removed from the ASC (a) upon having failed to have at least one representative attend two quarterly meetings in the prior twelve months, or (b) by vote of the other ASC members per the voting rules defined below. Attendance by a non-voting member of the organization (i.e., someone not officially designated as an alternate for voting purposes) counts towards meeting this requirement.
+
+The ASC meets quarterly, with at least one quarterly meeting each year to be held "in-person" at an agreed-upon location unless the ASC Members vote otherwise. A vote to "virtualize" the in-person meeting can only apply to that given year. Other quarterly meetings may be held by teleconference in accordance with the desires of ASC members. ASC quarterly meetings are conducted according to the following schedule:
+
```
dsolt: Do we need to clarify that can you still attend an in-person meeting virtually?   Any limitations on virtual attendance?

README.md
```diff
@@ -2,24 +2,251 @@
...
+
+## The Standardization Process
+
+Contributions of any form (e.g., use cases, questions, suggestions) are encouraged from anyone interested in the evolution of the PMIx Standard. The community relies on Issues and Pull Requests to discuss changes to the standard for a few reasons:
+
+* It provides an archive of the conversation around a change to the
```
dsolt: I would maybe avoid specifying Git concepts in this document.  I know we all love git and think it will be here forever, but this is like the 5th revision control system I've used and may not be the last.   Maybe the next one won't have pull requests.

README.md
rhc54: I believe we discussed this on the last call and agreed not to support virtual participation for in-person meetings as being too clumsy to really work. If the meeting is "in-person", then you have to physically be there to participate.

README.md
rhc54: Wow - I don't see any way to do that without creating total confusion. If the community migrates to some other mechanism in the future, you'll need to modify the governance to reflect the newly revised mechanism.

README.md
rhc54: We state that in the following paragraph, but I have no issue with restating it here as well:

> Informal participants do not have voting privileges on administrative matters, but are encouraged to attend the quarterly meetings and participate in those discussions. Individuals and/or organizations must attend meetings as informal participants in order to qualify for elevation to full Member status, as described below.

README.md
```diff
@@ -2,24 +2,250 @@
...
 
-## Contributing to the PMIx Standard
+## Governance
 
-There are two main ways to contribute to the PMIx Standard: Participating in the discussion, and writing sections of the document.
+The PMIx Standard is governed by an Administrative Steering Committee (ASC) comprised of representatives from a consortium of research, academic, and industrial organizations.The ASC shall consist of one representative from each Member organization, as defined below, and shall perform the following functions:
 
-### Participating in the discussion
+* Define the mission of the PMIx Standard.
+* Plan release timelines for versions of the PMIx Standard based on available resources, target features, and target goals of the Member organizations.
```
kathrynmohror: It is unclear what the target goals of the member organizations have to do with release time lines. As an example, a feature shouldn't be rushed just because a particular member organization has promised it to customers by a certain time -- the release cycles of the standard should be independent of the business of the implementations. It is also unclear on what 'available resources' means. Do you mean that release time lines could be altered, say slowed if there weren't enough people to make enough progress to warrant a release?

README.md
```diff
@@ -2,24 +2,250 @@
...
+* Vote on all issues requiring community consensus (e.g., acceptance of proposed modifications to the Standard), one vote per Member organization. In addition to voting on proposed modifications, the ASC can be requested to vote on any issue submitted by a Member for consideration.
```
kathrynmohror: It is a bit unclear what a Member is at this point. I think you mean that a Member is an organization, but it isn't clearly defined up front and then in the second sentence here when I am reading I wonder if a Member can be an individual person. It might be good to have a brief definition section up front for terms like Member, Organization, Participant, etc

README.md
```diff
@@ -2,24 +2,250 @@
...
+* Assemble working groups to address specific areas of interest to the community.
+* Nominate and approve the appointment of individuals to fill the role of Release Manager (per release), as described below.
 
-If you have a question about the PMIx interface or want to raise a question for discussion then please file an Issue against this repository. We are using the Issues to discuss topics, and will link Pull Requests related to that topic to that Issue. This provides the community with a public, archived forum for discussion.
+The ASC is led by two Co-Chairs responsible for organizing and conducting the quarterly meetings. This includes:
 
-### Participatin in the writing
+* Announcing meeting agendas at least one month in advance of the meeting. Any proposal wishing to be considered for the next meeting must be put forward (by assigning the proper label, as described below) prior to that time. The ASC Co-Chairs may allow a proposal to be added to the agenda after this time on a case-by-case basis at their sole discretion.
```
kathrynmohror: I'm not sure about the 'sole discretion' part since it could be seen as favoritism or some other kind of bias if some are allowed and others not. Perhaps there can be a semi-formal vote on proposals that are announced late. For example, before reading them there is a quick vote saying 'this proposal was late, is it okay for them to proceed with the reading?'.

README.md
```diff
@@ -2,24 +2,250 @@
...
+* Chairing the quarterly meetings and ensuring an opportunity for all to express their opinion during the meeting
```
kathrynmohror: I like the sentiment for 'ensuring an opportunity for all to express their opinion during the meeting' but I am not sure if it belongs in a governance document. What I mean is, the rest of the document aims to be clear cut about what the rules are but this is a soft requirement that is difficult/impossible to guarantee or measure. Perhaps it would go in a philosophy or code of conduct kind of document?

README.md
```diff
@@ -2,24 +2,250 @@
...
+* Maintaining and publishing minutes of the meeting, and tallying votes.
 
-If you want to write a portion of the standard (or even fix a typo or two), then use the Pull Request mechanism to post the change for review.
+Co-Chairs may vote on behalf of their organization subject to the one vote per Member organization limit. Co-Chairs serve for a period of two years, with seats filled on alternate years - i.e., one seat will be elected during the first meeting of each even year, and the other seat will be elected during the first meeting of each odd year. There is no restriction on the number of consecutive terms a given individual may serve - however, it is expected and desirable that the positions rotate amongst the Members.
```
kathrynmohror: What is the election process for the chairs? Meaning, is it like the voting rules described below or is it different?

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+* A Member is removed from the ASC (a) upon having failed to have at least one representative attend two quarterly meetings in the prior twelve months, or (b) by vote of the other ASC members per the voting rules defined below. Attendance by a non-voting member of the organization (i.e., someone not officially designated as an alternate for voting purposes) counts towards meeting this requirement.
+
+The ASC meets quarterly, with at least one quarterly meeting each year to be held "in-person" at an agreed-upon location unless the ASC Members vote otherwise. A vote to "virtualize" the in-person meeting can only apply to that given year. Other quarterly meetings may be held by teleconference in accordance with the desires of ASC members. ASC quarterly meetings are conducted according to the following schedule:
+
+* 3rd week of January
+* 3rd week of April
+* 3rd week of July
+* 3rd week of October
```
kathrynmohror: I'd argue that specifying the week  of the meeting (even though it is allowed to change) is too specific. I'd recommend saying something more general like 'once per calendar quarter with a preference for the first month of the quarter' (or with better wording than that)

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+ASC Members may, if desired, vote to alter the timing of the meetings in a given year to accommodate Member schedules, subject to the voting rules described below. Changes to a given meeting's dates must be determined no later than the prior quarterly meeting.
```
kathrynmohror: If the meeting is face to face, I would give longer lead time than a quarter for changing the meeting date since travel (especially international travel) can require a long time to get approval for in some cases.

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+The Co-Chairs shall include links to each proposed modification to the Standard up for consideration during that meeting in the published agenda.
+
+#### Current ASC Members
+
```
kathrynmohror: I don't think current membership belongs in the governance document so that it doesn't have to be versioned each time membership changes. Perhaps the document could refer to a web page or something

README.md
```diff
@@ -2,24 +2,250 @@
...
+<div  align="center">
+<table>
```
kathrynmohror: Are these just examples of who could be members? How do we bootstrap this?

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+* Each active ASC Member organization will receive one vote. All votes are required to pass with 2/3 majority. A quorum (>50%) of Members must be present (either physically or virtually). Abstentions will not be counted in the total votes when calculating whether there is a 2/3 majority.
+
+* A Member may submit a vote _in absentia_ to the ASC Co-Chairs in advance of the meeting if attendance by a representative is not possible (e.g., a quarterly meeting that falls during a national holiday period), but each ASC Member organization can only cast one vote. This prevents a Member from voting on behalf of its own entity and acting as a proxy for another member. _In absentia_ votes are counted by the Co-Chairs as if the member were present (thus counting towards the meeting's quorum requirement), but are recorded as having been voted _in absentia_ in the official tally.
+
+Formal votes may, if requested by at least one Member, be conducted as "secret ballots" to abate concerns over customer and/or competitor perceptions. All votes are tallied by the Co-Chairs and recorded in the minutes of the meeting.
```
kathrynmohror: The secret ballot seems strange to me in a public forum like this. I think this point needs discussion

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+* A Member is removed from the ASC (a) upon having failed to have at least one representative attend two quarterly meetings in the prior twelve months, or (b) by vote of the other ASC members per the voting rules defined below. Attendance by a non-voting member of the organization (i.e., someone not officially designated as an alternate for voting purposes) counts towards meeting this requirement.
+
+The ASC meets quarterly, with at least one quarterly meeting each year to be held "in-person" at an agreed-upon location unless the ASC Members vote otherwise. A vote to "virtualize" the in-person meeting can only apply to that given year. Other quarterly meetings may be held by teleconference in accordance with the desires of ASC members. ASC quarterly meetings are conducted according to the following schedule:
+
```
kathrynmohror: Can any participant attend the face to face meeting or is it only able to be attended by the ASC?

README.md
kathrynmohror: Perhaps it could be accomplished by separating out the general steps for standardization from the actual implementation of them. What I mean is, you could talk in general terms about creating a proposal that is in a "draft" state and how that proposal moves from draft to reading to provisional, etc. Then you can refer to a "workflow document" or something like that that outlines how we would use github to manage the proposals. Then later when github is replaced by something else, the changes would be compartmentalized. 

All that said, I am not sure which approach is better. You need to tell people exactly what to do somewhere...

README.md
```diff
@@ -2,24 +2,250 @@
...
+Contributions of any form (e.g., use cases, questions, suggestions) are encouraged from anyone interested in the evolution of the PMIx Standard. The community relies on Issues and Pull Requests to discuss changes to the standard for a few reasons:
+
+* It provides an archive of the conversation around a change to the
+standard that can be vital in understanding the rationale for the change if that conversation
+needs to be revisited in the future.
+
+* It facilitates asynchronous conversation from
+a geographically distributed user-base. To that end, any discussion on the teleconference
+about specific Issues or Pull Requests will be noted as a comment to that Issue or Pull Request.
```
kathrynmohror: This is possibly contradicted  by what is in the paragraph on line 160 ("The Standardization Process remains essentially...") where it says notes will be in the wiki. Do the notes have to be also included in the github issues?

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+Note that a proposal can be withdrawn by the author(s) at any time for any reason up to the time the proposal is committed to the Standard, assuming acceptance.
+
+#### "Classes" of Standard
+The PMIx Standard is composed of a combination of APIs and key-value attributes. APIs are intentionally designed to be "lightweight" - i.e., the API signature is generic with a minimal number of fixed parameters in its signature. With few exceptions, APIs are required to include an array of key-value attributes in their signature to facilitate behavioral extensions without modifying the API signature. Thus, PMIx attempts to minimize deprecation and modification of APIs as use models evolve over time.
+
+Each API can designate a given attribute as either "**required**" or "**optional**" - this indicates whether or not a PMIx implementation of that API is required to support the attribute. PMIx implementations are requested to support optional attributes, but are not required to do so. This is intended to both provide flexibility to PMIx implementors (e.g., to tailor the implementation to their target audience) and to ease adoption of the Standard.
```
kathrynmohror: These two paragraphs about required or optional APIs or attributes seem out of place. I think they belong in the standard document but not in this procedure document. What I mean is the ASC won't vote to decide if an attribute/API is required or optional, that would be part of the proposal

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+    All proposed additions to the PMIx Standard enter the PMIx Standard under Provisional status. Attributes in the "Provisional" class may only be designated as "optional" by an API - i.e., an API definition cannot designate a "Provisional" attribute as "required". Users, however, are always free to "require" that an attribute be supported when calling the API, thus directing the API's response should the attribute not be supported by it.
```
kathrynmohror: I think I understand what you are going for here, but it is a bit confusing since there is so much flexibility in what should be implemented in the standard already. What if the provisional addition is meant to be a required interface/attribute in the stable version in order for some functionality to work? How can one try it out properly without knowing the required/optional parts of the eventual stable interface?

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+Anyone indicating concern must provide a comment on the ticket following the "**Straw Poll**" comment describing their concern. Sufficient detail should be provided to allow the author(s) to respond to the specific concerns. The author(s) can then work to address the concern, modifying the PR as required - modification should be accompanied by an appropriate comment citing the original participant who raised the concern to ensure they are notified of the change. Participants shall indicate their acceptance or rejection of the modification by so marking the change comment and adjusting their original "vote" on the straw poll if appropriate.
+
+The ASC Co-Chairs shall monitor the status of discussion on the PR. Once discussion appears to have consolidated (either measured by participation in the straw poll, rate of comments, or a combination of both), the Co-Chairs will schedule the proposal for a reading at the next quarterly meeting of the ASC. The proposal shall be included in the announced agenda for the meeting along with a link to the PR and the results of the straw poll.
```
kathrynmohror: I'd also advocate for the proposal authors to have an active role here. If the control to bring forward a proposal is only under the discretion of the chairs (and the chairs say don't pay attention or whatever), progress could be stalled. Can authors request to schedule a reading? Is there a reason why the chairs would deny the request? (assuming the rest of the procedures have been followed)

README.md
```diff
@@ -2,24 +2,250 @@
...
+
+At least one proposal author must attend the quarterly meeting to present the PR and participate in discussion regarding its approval. A "reading" of the proposal shall be conducted at the meeting with the author(s) reviewing any questions/concerns raised and how they were addressed. During the presentation, someone designated by the Co-Chairs will take notes on the PR (and/or Issue) on behalf of the presenter. At the end of the agenda item, those members wishing to adjust their straw votes will be given an opportunity to do so. Members who did not attend the meeting will also have a chance to adjust their votes based on the published minutes of the meeting.
```
kathrynmohror: Can participants adjust their straw votes too based on the discussion or only members?

rhc54: Anyone can:

> At the end of the agenda item, anyone wishing to adjust their straw votes will be given an opportunity to do so. Participants who did not attend the meeting will also have a chance to adjust their votes based on the published minutes of the meeting.

README.md
rhc54: > Once discussion appears to have consolidated (either measured by participation in the straw poll, rate of comments, or a combination of both), or at the request of the proposal's author(s), the Co-Chairs will schedule the proposal for a reading at the next quarterly meeting of the ASC.

README.md
rhc54: > Attributes that are intended to become designated as "required" by an API when elevated to "Stable" status shall be clearly designated as such in the proposal.

README.md
rhc54: I guess I can see it both ways - folks requested it for clarification of the rest of the language in the governance writeup. Note that we don't really have a "document" here. The governance description is just a readme at this stage.

README.md
rhc54: Good catch:

> Discussions regarding a proposal are conducted either on the Issue/PR itself or during the regular community teleconference (and captured as notes on the Issue/PR). Teleconferences provide a high bandwidth discussion format and valuable synchronization mechanism to guide progress. Non-issue/PR related notes from the teleconferences are archived on the PMIx community's [wiki](https://github.com/pmix/pmix-standard/wiki#meeting-information).

README.md
rhc54: We actually do state in the paragraph above this that anyone is encouraged to attend the meetings. However, I can reiterate it here.

> Non-ASC participants are welcome to attend any meeting in a non-voting capacity.

README.md
rhc54: Sure - it is intended to help vendors when confronted by something a customer (either current or potential) really wants but the vendor opposes.

README.md
rhc54: These are all organizations that have been active members of the "informal" ASC over the last few years - they are the "seed" for the updated organization as they already have met the "member" requirements.

README.md
rhc54: I believe the governance "doc" is just the readme on the standards repo, so there is no real versioning involved here. I have no strong feelings about it, but the doc does say that members are given visible acknowledgement of their participation, and this seems a relatively simple place to provide it.

README.md
rhc54: True - I can adjust it to two quarters for now. Difficult to know how far in advance to make it as everyone's time requirements are different.

> Changes to a given meeting's dates must be determined no later than two quarterly meetings before the one being changed.

README.md
rhc54: Good point:

> ASC quarterly meetings are conducted once per calendar quarter with a preference for a time during the first month of the quarter. Precise meeting dates are determined by the ASC members in accordance with their schedules, subject to the voting rules described below. Subsequent changes to a given meeting's dates must be determined no later than two quarterly meetings before the one being changed.

README.md
rhc54: > Co-Chairs are elected from amongst the ASC members during the first meeting of each calendar year (using the voting system described below) and serve for a period of two years, with seats filled on alternate years - i.e., one seat will be elected during the first meeting of each even year, and the other seat will be elected during the first meeting of each odd year.

README.md
rhc54: Yeah, I can somewhat picture what you are saying - I'm just concerned that it gets too fragmented. In other words, someone has to read this writeup, then go read the workflow writeup, then try to piece it all together. Seems a little imposing and prone to lots of "but how do I" questions.

README.md
rhc54: The intent here is to capture a little reality in terms of how the ASC works. There is no point in defining a target release date if there aren't people available to review things and move them along. Witness the lengthy delays in getting other standards out.

Similarly, the reality is that releases of the standard are always going to be driven by the target goals and features of the Members. They are the ones voting, and they are primarily the ones proposing new features. As a result, the release timeline is largely going to be driven by their business needs, not by arbitrary dates.

So while in principle things should be separate, the reality is that things tend to be rather coupled. Nothing different from what we see in other standard bodies. This bullet just states that one function of the ASC is to propose and manage release timelines that reflect the needs of its members.

README.md
rhc54: > The ASC shall consist of one representative from each Member organization, **as defined below**

>  Note: Individuals may also become Members so long as their direct employer is not a current Member of the ASC.

It's a bit of a chicken/egg problem - if you start out by defining Member, then the immediate question is "member of what?". I can try to be a little clearer early on...

README.md
rhc54: Problematic as it doesn't give any warning and might cause a particular individual to miss a meeting that discussed something of importance to them. It also doesn't give members a chance to even look at the proposal prior to the meeting. Maybe the right answer is just to be hard about it and say "no" to anything that is late?

README.md
rhc54: I agree that it isn't a governance rule - just a loose set of definitions of the Co-Chair duties. I'd rather limit the number of documents describing all this so people can just go to one place to get what they need. If it feels too squishy, then I'd suggest we just remove that phrase.

README.md
```diff
@@ -2,24 +2,243 @@
...
+
+#### Provisional acceptance
+The process for acceptance to Provisional status begins when the author(s) assign an "**RFC**" label to the GitHub PR, add an empty "**Straw Poll**" comment to it, and send an announcement of the PR to the PMIx community mailing list requesting consideration of the PR for Provisional status. The announcement shall include a link to the PR, a brief description of the proposed change, and indicate any desired timeline for approval. It is especially important that proposals tied to contractual milestones be so designated to ensure that members are alerted to potential time pressures.
+
+The "Straw Poll" comment is used by the ASC as a mechanism for obtaining a sense of how many people have viewed the PR and are following the discussion. This serves no official purpose other than to help define participation in the consensus building process and record the following tally:
```
jjhursey: With the new "Straw Poll" above, maybe we just do a back reference to that paragraph for guidance on the comment to post and the various emojis.

README.md
SteVwonder: If it would be helpful to potential contributors, we can add a `CONTRIBUTING` document that summarizes the contributing steps (with lots of details about Github) and elides the governance details.

README.md
```diff
@@ -2,24 +2,243 @@
...
+
+Upon successfully completing the straw poll, the Co-Chairs shall label the PR as "**Eligible**" to indicate that the proposal is eligible for formal consideration by the ASC. Given the significant commitment conferred by the PMIx community to "Stable" elements, formal approval of elevation requires that the ASC conduct a full presentation on the proposal at two consecutive quarterly meetings. At the conclusion of the first meeting, a formal vote of the ASC shall be held (per the above voting rules) to determine if the proposal merits further consideration. The Co-Chairs shall set a GitHub label indicating the result as either "**First Pass**" or "**Pushed Back**".
```
jjhursey: Clarify “First Pass” means “First Vote Passed” - maybe just adjust the label name to make it clear.

jjhursey:
I captured some notes from the teleconf today in the [meeting notes](https://github.com/pmix/pmix-standard/wiki/Meeting-2019-07-12) on the wiki.

The general sense of the room is that this PR is a good path and nearly complete. For those that contributed comments please indicate those that have been resolved so we can focus on closing those unresolved discussion items.

Hope is that we stabilize this PR in the next week. Then during the following teleconf we can start the formal process of acceptance.

README.md
rhc54: Done

README.md
rhc54: Done

rhc54:
I have updated the text per the comments from the telecon of 7/12/2019. Please take a look when you can!

README.md
```diff
@@ -2,24 +2,280 @@
...
+Co-Chairs are elected from amongst the ASC members during the first meeting of each calendar year (using the voting system described below) and serve for a period of two years, with seats filled on alternate years - i.e., one seat will be elected during the first meeting of each even year, and the other seat will be elected during the first meeting of each odd year. There is no restriction on the number of consecutive terms a given individual may serve - however, it is expected and desirable that the positions rotate amongst the Members. Likewise, there is no requirement that a Co-Chair be the Representative or Alternate for their Member. Co-Chairs serving as Representative or Alternate for a Member may vote on behalf their behalf subject to the one vote per Member limit.
```
abouteiller: typo: behalf their behalf

README.md
```diff
@@ -2,24 +2,280 @@
...
+This is not a binding majority-rule vote, but it will be a very
+significant input into the corresponding ASC decision.
+
+Here are the meanings for the emojis:
+
+    * Celebration / Hooray / Tada!: I support this so strongly that I
+          want to be an advocate for it
+    * Heart: I think this is an ideal solution
+    * Smiley face: I'd be happy with this solution
+    * Thumbs up: I'd tolerate this solution
```
abouteiller: Recommend using `thumb up` 👍  as a synonym to `smiley face` 😃 
and use  `:neutral_face` 😐 for the 'tolerate this solution'

README.md
```diff
@@ -2,24 +2,280 @@
...
+
+* "**Deprecated**" entries are slated for removal in an upcoming major release of the Standard. Deprecation warning must be provided for a minimum of one major release - i.e., an entry that is marked for deprecation in release A.0 cannot be removed from the Standard until at least the A+2.0 release. Extensions to the deprecation deadline can be granted on a case-by-case basis upon vote of the ASC.
```
abouteiller: What happens to cross-version compatibility should be discussed for deprecated API/keys.

README.md
```diff
@@ -2,24 +2,280 @@
...
+
+The final straw vote results shall be tallied no later than two weeks following the quarterly meeting where the proposal was presented. At that time, the Co-Chairs shall set a GitHub Label corresponding to the result - either indicating "Accepted" or "Pushed Back" based on the final results of the "**Straw Poll**". Accepted proposals can be committed to the next minor release version by the Release Manager in accordance with their schedule. Proposals that are pushed back for further work can be brought forward again at a later date, or can be withdrawn by the author(s) by simply closing the PR.
+
+#### Provisional process flow
```
abouteiller: Process flow for provisional proposals

README.md
rhc54: Done

README.md
rhc54: Done

README.md
rhc54: Added to deprecation section:

> Impact on backward or cross-version compatibility must be identified in the discussion to ensure adequate consideration is given to such issues.

README.md
rhc54: done

README.md
```diff
@@ -2,24 +2,280 @@
...
+High Performance Computing (HPC) libraries and programming models with
+portable and well defined access to commonly available
+distributed computing system services.
 
-## Contributing to the PMIx Standard
+## Governance
 
-There are two main ways to contribute to the PMIx Standard: Participating in the discussion, and writing sections of the document.
+The PMIx Standard is governed by an Administrative Steering Committee (ASC) comprised of representatives from a wide range of research, academic, and industrial organizations. Membership in the ASC is determined by participation in the PMIx Standard meetings and a vote of the ASC. Anyone (individual or organization) meeting the membership criteria can become an ASC Member. Thus, the term "Member" may refer to either an individual or an organization (e.g., corporation, university, or national laboratory). Each Member shall be represented by a designative individual (the "Representative") who is authorized to vote on behalf of the Member. If desired, a Member may designate an "Alternate" who is authorized to vote on behalf of the Member on occasions when the Representative is not present.
```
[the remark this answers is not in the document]
rhc54: got it - thanks!

rhc54:
Dear all:

I have updated the document based on everyone's comments and squashed the results to allow for clean review of its current state. The only significant changes from this morning were:

* added election of two Secretaries to take minutes of meetings and maintain the web site - elections mirror the Co-Chairs
* moved the election to the last meeting of each calendar year, with new officers taking their positions at the beginning of the upcoming year
* modified the emoji list for the straw poll. Turns out that GitHub has restricted the number and type of the emojis one can assign to a comment.
* cleaned up some language per people's comments

Please carefully review the final doc. As per the telecon, I have attached a "straw poll" for your use, and we intend to take a formal vote on the governance document during the telecon of Aug 2nd

rhc54:

          we know your current thoughts

          want to be an advocate for it
          other technologies instead

jjhursey:
Ref https://github.com/pmix/governance/pull/1 (LaTeX translated version)

jjhursey:
Vote **Passed** on Aug. 2, 2019 Teleconf (recorded in the [meeting minutes](https://github.com/pmix/pmix-standard/wiki/Meeting-2019-08-02) ):
```
Yes:
 - Intel (sent via email to Co-Chair)
 - Argonne National Laboratory (sent via email to Co-Chair)
 - IBM
 - Altair
 - LLNL
 - ORNL
 - Microsoft
 - Cray
 - TU Munich
 - Sylabs
 - UTK

No:
 - 

Abstain:
 - 
```

PR #196, opened 2019-08-02
Removes governance text from README file 

kathrynmohror:
The purpose of this commit is to remove the governance rules text from the README.md file in the pmix-standard repo and instead point interested readers to the pmix.org web site where the governance rules and other information (e.g. about joining meetings) will be posted.


kathrynmohror:

          we know your current thoughts

          want to be an advocate for it
          other technologies instead

Issue #194, opened 2019-07-31
Definition of `PMIX_GROUP_ASSIGN_CONTEXT_ID`


SteVwonder:
`PMIX_GROUP_ASSIGN_CONTEXT_ID` and `PMIX_GROUP_NOTIFY_TERMINATION` are both defined to be "Notify remaining members when another member terminates without first leaving the group. The default is false".  I assume this definition is correct for the latter and the former got the definition through copy/paste.

I'm not sure what is the intended definition of `PMIX_GROUP_ASSIGN_CONTEXT_ID`.  There isn't any text that I am able to find in the standard or RFCs that describes this attribute or the (presumably related) `PMIX_GROUP_CONTEXT_ID` attribute.

### References

This primarily applies to Chapter 3 (where the attribute is defined), but it affects Chapter 12 as well (where the attribute is referenced).

rhc54:
Please see #197 for clarification. The primary use-case was for the MPI Session folks as a means of creating a globally unique communicator ID at time of group construct, thus avoiding another operation to compute one. However, there is nothing in PMIx that constrains the context ID to that usage.

PR #197, opened 2019-08-05
Clarify the Group Context ID references

### LINKED ISSUE #195: Clarification on `PMIX_GROUP_FT_COLLECTIVE`
`PMIX_GROUP_FT_COLLECTIVE` is defined in Chapter 3 to mean "Adjust internal tracking for terminated processes. Default is false"

The only reference to `PMIX_GROUP_FT_COLLECTIVE` that I can find is under `PMIx_Group_leave` in Chapter 12.  The text there states: "All PMIx-based collectives such as `PMIx_Fence` in action across the group will automatically be adjusted if the collective was called with the `PMIX_GROUP_FT_COLLECTIVE` attribute (default is false) – otherwise, the standard error return behavior for that collective will be executed."

### Suggested Clarification

Do any of the collectives currently support this attribute? If so, I think it would be useful to include `PMIX_GROUP_FT_COLLECTIVE` in the `Format` sections of the relevant collectives.  If none of them support it, it may make sense to remove it from the standard until one does.

### References

Chapter 3 (attribute definition) and Chapter 12 (attribute reference).

rhc54:
Fix a copy/paste error. Explain what a context ID is and where it comes
from, and give an example of its use.

Refs #194

Remove the PMIX_GROUP_FT_COLLECTIVE attribute
Pending further interest/support from the FT community
    
Fixes #195


SteVwonder:
Thanks @rhc54!

PR #202, opened 2019-08-08
Add chapter on scheduler interfaces

rhc54:
The PMIx server library includes several interfaces specifically intended to support schedulers by providing access to information of potential use to scheduling algorithms - e.g., information on communication costs between different points on the fabric. These interfaces are restricted in availability to the scheduler due to their high cost, both in terms of execution and memory consumption.


rhc54:
@pmrich @dsikich 

Could you (and your team) perhaps take a look at this and see if it matches expectations?

rhc54:

          we know your current thoughts

          want to be an advocate for it
          other technologies instead

dholmes-epcc-ed-ac-uk:
What is the definition of a scheduler w.r.t. to the restriction of availability of some parts of the server interface? How does a scheduler identify itself as a scheduler to the server?

A brief look at the changes in this PR suggests that access is gated by a register/de-register API, which means that *any* component could be a "scheduler".

Maybe strike "to the scheduler" or replace it appropriately:
> These interfaces are restricted in availability (via explicit registration) due to their high cost,

rhc54:
@dholmes-epcc-ed-ac-uk It is a little more subtle than that, and so I have tried to provide more verbage explaining the "restriction", and included the missing attribute definition that goes with it.

dholmes-epcc-ed-ac-uk:
Thanks @rhc54, that definitely makes things clearer.

Chap_API_Scheduler.tex
```diff
@@ -0,0 +1,237 @@
...
+\pastePRIAttributeItem{PMIX_NETWORK_PLANE}
+
+\reqattrend
+
+%%%%
+\descr
+
+Register for access to fabric-related information, including communication cost matrix. This call must be made prior to requesting information from a fabric. The caller may request access to a particular \refterm{network plane} via the \refattr{PMIX_NETWORK_PLANE} attribute - otherwise, the default fabric will be returned.
+
+If available, the \refarg{fabric} struct will contain the address and size of the communication cost matrix associated with the specified network plane. For performance reasons, the \ac{PMIx} server library will \emph{not} provide thread protection for cost matrix access. Instead, users are required to register for \refconst{PMIX_FABRIC_UPDATE_PENDING} events indicating that an update to the cost matrix is pending. When received, users are required to terminate any actions involving access to the cost matrix before returning from the event.
```
gvallee:
```suggestion
If available, the \refarg{fabric} struct shall contain the address and size of the communication cost matrix associated with the specified network plane. For performance reasons, the \ac{PMIx} server library does \emph{not} provide thread protection for cost matrix access. Instead, users are required to register for \refconst{PMIX_FABRIC_UPDATE_PENDING} events indicating that an update to the cost matrix is pending. When received, users are required to terminate any actions involving access to the cost matrix before returning from the event.
```

Chap_API_Scheduler.tex
```diff
@@ -0,0 +1,237 @@
...
+\begin{arglist}
+\argin{input}{address of a \refstruct{pmix_fabric_t} (handle)}
+\end{arglist}
+
+Returns \refconst{PMIX_SUCCESS} or a negative value corresponding to a \ac{PMIx} error constant.
+
+%%%%
+\descr
+
+Deregister a fabric object, providing an opportunity for the \ac{PMIx} server library to cleanup any information (e.g., cost matrix) associated with it
```
gvallee:
```suggestion
Deregister a fabric object, providing an opportunity for the \ac{PMIx} server library to cleanup any information (e.g., cost matrix) associated with it.
```

Chap_API_Scheduler.tex
```diff
@@ -0,0 +1,237 @@
...
+\end{arglist}
+
+%%%%
+\descr
+
+Returns one of the following:
+
+\begin{itemize}
+    \item \refconst{PMIX_SUCCESS}, indicating that the request is being processed by the host environment - result will be returned in the provided \refarg{cbfunc}. Note that the library must not invoke the callback function prior to returning from the \ac{API}.
+    \item \refconst{PMIX_OPERATION_SUCCEEDED}, indicating that the request was immediately processed and returned \textit{success} - the \refarg{cbfunc} will not be called
```
gvallee:
```suggestion
    \item \refconst{PMIX_OPERATION_SUCCEEDED}, indicating that the request was immediately processed and returned \textit{success} - the \refarg{cbfunc} will not be called.
```

Chap_API_Scheduler.tex
```diff
@@ -0,0 +1,237 @@
...
+    \item a \ac{PMIx} error constant indicating either an error in the input or that the request was immediately processed and failed - the \refarg{cbfunc} will not be called
```
gvallee:
```suggestion
    \item a \ac{PMIx} error constant indicating either an error in the input or that the request was immediately processed and failed - the \refarg{cbfunc} will not be called.
```

gvallee:
I do not remember all the rules and could not find them while looking for them for a minute, so I hope it is okay to suggest modifications directly in the PR. If not all my apologies.

rhc54:
> I do not remember all the rules and could not find them while looking for them for a minute, so I hope it is okay to suggest modifications directly in the PR. If not all my apologies.

No problem - this is fine. I'll eventually have to squash them as GitHub doesn't add signatures to the suggestions.

jjhursey:
Is there an implementation branch associated with this PR?

rhc54:
Hmmm...it actually is already committed to our master branch, here are the primary relevant commits:

pmix/pmix@c64c4f507
pmix/pmix@31805b082
pmix/pmix@42a2d2179
pmix/pmix@a672eac1b

pmrich:
We noticed that in the documentation for PMIx_server_get_index there is a reference to a cbfunc, however we aren't finding any other reference to the callback being defined or passed in in these functions in this section.

rhc54:
🤣 Copy-paste error! Fixed

PR #209, opened 2019-08-19
Add blocking form of PMIx_Query_info

rhc54:
Identical to existing non-blocking form except for function signature

### DIFF
--- a/Chap_API_Job_Mgmt.tex
+++ b/Chap_API_Job_Mgmt.tex
@@ -97,6 +97,121 @@ The caller is responsible for releasing the string when done with it.
 
 
 %%%%%%%%%%%
+\subsection{\code{PMIx_Query_info}}
+\declareapi{PMIx_Query_info}
+
+%%%%
+\summary
+
+Query information about the system in general.
+
+%%%%
+\format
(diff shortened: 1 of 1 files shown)


rhc54:

          we know your current thoughts

          want to be an advocate for it
          other technologies instead

rhc54:
We could - however, that isn't what we did for all the other blocking/non-blocking pairs. I simply copied what we had done. I agree, though, that it seems awfully duplicating.

IIRC, I think the reasoning was that each function should "stand alone" in the standard so that if someone uses the TOC to jump to the non-blocking form, everything they need is right there.

I'd suggest we let the "rewrite" WG decide which way they want to do it, and if they want to remove the duplication, then do so for all functions at the same time.

jjhursey:
You are correct. I thought we just left them as back-references, but looking at `PMIx_Get`/`PMIx_Get_nb` I see we duplicate the attribute list and description.

I agree then that we should duplicate the text. Someone can go through and de-duplicate the text in a PR if they wish. I don't think that falls under any existing working group, and probably doesn't need to if someone is interested to do it.

PR #210, opened 2019-08-19
Support return of static values for PMIx_Get

rhc54:
Provide an attribute indicating that the user requests that any
pointers in the returned value point directly to values in the
key-value store. Define a return status of PMIX_ERR_GET_MALLOC_REQD
to indicate that direct pointers could not be supported - thus,
the returned data contains allocated memory that the user must release.

If PMIX_GET_STATIC_VALUES is provided to the blocking PMIx_Get, then
assume that the provided address points to a statically defined memory
location. Returned non-pointer values shall be copied directly into the
provided memory - i.e., do not malloc a pmix_value_t to contain the
returned values. User takes responsibility for not releasing memory
unless the "malloc_reqd" status is returned.


rhc54:
@raffenet Please take a look

rhc54:

          we know your current thoughts

          want to be an advocate for it
          other technologies instead

jjhursey:

rhc54:
I'm going to hold off on this one pending completion of an implementation, just in case something unexpected appears.

PR #212, opened 2019-08-24
Add fields and support macros to pmix_coord_t definition

rhc54:
Add fields to help identify the fabric and plane to which the coordinate
values refer. Provide support macros for constructing and releasing the
pmix_coord_t fields.


rhc54:

          we know your current thoughts

          want to be an advocate for it
          other technologies instead

jjhursey:

PR #213, opened 2019-08-24
Restore the appendix on support macros

### LINKED ISSUE #178: Argv and other macros missing
Some of the macros were moved to an appendix that was subsequently commented out in the document. Should they be restored? Also seems that a few we defined in PRI and are commonly used are missing.

rhc54:
Add in missing macros. These should probably be added to the standard
itself (instead of in an appendix) and indicate that an implementation
may not provide them. Note that current users are taking advantage of
these and so there may be some portability issues to consider.

Fixes #178


jjhursey:
I agree that these are frequently used enough that we will likely want them the v4 standard for reference. The chapter does call them out as being specific to the PRI which is good.

rhc54:
I think I'm arguing that these macros should not be called out as PRI-specific and put in an Appendix. Instead, they should simply be defined in the PMIx_Struct chapter and left to the implementation to provide however they choose to do so - in the hope that user code that utilizes those macros can be portable.

jjhursey:
I could probably be convinced of that. I think my hesitation is in standardizing preprocessor macros which feels a bit odd - but that's not a good reason.

rhc54:
@jjhursey I moved the macros to "first-class" status as we have a number of similar macros already in the standard - there doesn't seem to be a good explanation as to why these are being singled out as something different. See what you think.

jjhursey:

Issue #185, opened 2019-05-08
Many attributes presented in request-related attributes sectiion have no clear use


dsolt:
I see several attributes such as:

PMIX_NOTIFY_COMPLETION 
PMIX_JOB_TERM_STATUS 
PMIX_ANL_MAP
(and I believe many more)

These are presented in a very generic way, but no PMIx routine is given that accepts these as required or optional.   If the goal it to reserve these for some internal use or some other non-documented purpose, then we should note that somewhere.  As it is now, they are just kind of "there" with no real explanation of how they would get used.

rhc54:
I'll update the doc and try to capture these. Some of them relate to tools which will be covered in the upcoming chapter. All are actually used in user interactions, not internally.

rhc54:
@jjhursey Perhaps it would be possible to provide a report that shows attributes that are not cited by any API? All the attributes are actually used somewhere - it could be that people just aren't seeing the reference (most likely reason) or that the reference is missing, or it is coming in a future chapter. Without seeing which ones are not referenced, it becomes almost impossible to chase them down.

jjhursey:
Yeah, I think that should be possible. It'll require a bit of LaTeX hackery to get right. I'll put it on my to do list and try to tackle it this weekend.

rhc54:
@jjhursey Let me know if you get a chance to do this.

jjhursey:
@rhc54 I posted a script to do the attribute check (and it's current output) in PR #223. Let me know what you think.

jjhursey:
We added a check script (ran at the bottom of 'make') that makes sure that for every attribute declared it is connected to some usage in the text. I think going beyond just a reference is a topic for the Slices/Grouping functionality working group - meaning make a proper use case for each one.

```
make check-attr-ref
====> Checking for Attributes Declared, but not referenced
  0 of 462 Attributes are missing reference
```

PR #214, opened 2019-08-25
Clarifications

### LINKED ISSUE #170: Clarify language around PMIx_Connect
Clarify that PMIx_Connect does _not_ require the exchange of job-level info nor inclusion of modex info. The exchange of job-level info should be noted as an "advice to implementers" as a possible optimization. The inclusion of modex info isn't possible as there is no way for the host environment to deliver that info to the PMIx server.

### LINKED ISSUE #184: Clarify intent of cbdata in function signatures
In a discussion earlier today it was pointed out that the PMIx standard is unclear on the intent of the `cbdata` option to many of the nonblocking interfaces. This should be clarified in the standard.

Below is a rough explanation that probably needs more polish:

> Many PMIx interfaces, particularly nonblocking interfaces, use a `void* cbdata` object passed to the function that is then passed to the associated callback. In a client-side API, the `cbdata` is a client-provided context (opaque object) that the client can pass to the nonblocking call (e.g., `PMIx_Get_nb`). When the nonblocking call (e.g., `pmix_value_cbfunc_t`) completes the `cbdata` is passed back to the client _without modification_ to allow the client to associate a context with that callback. This is useful if there are many outstanding nonblocking calls. The additional arguments to the callback function convey any additional information (e.g., `kv`) requested in that operation. 
>
> A similar model is used on the server-side API, though slightly inverted as the PMIx library is often the one providing the `cbfunc` and `cbdata` to be used when the requested operation (e.g., fence) is complete. The PMIx library is making an upcall into the RM via the PMIx server-side API with a specific `cbfunc` and `cbdata`. The PMIx library expects the server to call the `cbfunc` with the necessary arguments and pass back the original `cbdata`. This gives the server-side PMIx library the ability to associate a context with the call back (since multiple operations may be outstanding). The `cbdata` that the server sees is most likely not the same as what the client put into their side of the interface, but likely context specific to the PMIx implementation.

### LINKED ISSUE #186: PMIX_JOB_CTRL_ID  has no clear use
A host is required to accept this , but the only description is "Provide a string identifier for this request."   What is a host expected to do with it?

### LINKED ISSUE #208: Clarify example text around key/value management regarding PMIX_NUM_NODES
Chapter 5 Line 7:

> Keys are required to be unique within a specific level of informarion as defined in 3.4.11. For example, a value for PMIX_NUM_NODES can be specified for each of the session , job , and application levels. However, subsequently specifying another value for that attribute in the session level will overwrite the prior value.

Besides the typo (informarion), This may be a bad example since the user is not allowed to change PMIX_NUM_NODES based on this advice to users on the next page:

Note that keys starting with a string of “pmix” are exclusively reserved for the PMIx standard and must not be used in calls to PMIx_Put . Thus, applications should never use a defined “PMIX_” attribute as the key in a call to PMIx_Put .

This leads to confusion as to what the text around PMIX_NUM_NODES is trying to communicate.   I think the example is trying to explain that the resource manager could 
change the value of PMIX_NUM_NODES at the session level when a new job was added, for example.   If that is the case, then I propose we use the text below, however I also would note that I don't believe the PMIx reference implementation has any way to do this currently which either makes this a poor example or calls for changes in the PRI.   (How would a server using the PMIx reference implementation update PMIX_NUM_NODES as the session level? ).   

### Suggested Clarification

Keys are required to be unique within a specific level of information as defined in 3.4.11. For example, a value for PMIX_NUM_NODES can be established by the PMIx server for each of the session, job , and application levels.  If new nodes were introduced to a session through a  PMIx_Spawn call, the server could change the value in the session level which would change the value at the session level for all clients running inside that session.   

### References

Chapter 5 line 7.

rhc54:
Clarify use of PMIX_JOB_CTRL_ID
Fixes #186

Remove confusing paragraph at start of key-value mgmt
Fixes #208

Clarify intent of cbdata in function signatures
Fixes #184

Clarify language around PMIx_Connect
Fixes #170

 Reference some attributes where they can be used
 Refs #185

### DIFF
--- a/Chap_API_Struct.tex
+++ b/Chap_API_Struct.tex
@@ -3699,7 +3699,7 @@ This attribute would direct \refapi{PMIx_Finalize} to execute a barrier as part 
 
 %
 \declareAttribute{PMIX_JOB_TERM_STATUS}{"pmix.job.term.status"}{pmix_status_t}{
-Status returned by job upon its termination.
+Status returned by job upon its termination. The status will be communicated as part of a \ac{PMIx} event payload provided by the host environment upon termination of a job. Note that generation of the \refconst{PMIX_ERR_JOB_TERMINATED} event is optional and host environments may choose to provide it only upon request.
 }
 
 %
@@ -3707,6 +3707,9 @@ Status returned by job upon its termination.
 State of the specified process as of the last report - may not be the actual current state based on update rate.
 }
 
+\declareAttribute{PMIX_PROC_TERM_STATUS}{"pmix.proc.term.status"}{pmix_status_t}{
+Status returned by a process upon its termination. The status will be communicated as part of a \ac{PMIx} event payload provided by the host environment upon termination of a process. Note that generation of the \refconst{PMIX_PROC_TERMINATED} event is optional and host environments may choose to provide it only upon request.
+}
 
 %%%%%%%%%%%
 \subsection{Server-to-PMIx library attributes}
@@ -4541,7 +4544,7 @@ Resume (``un-pause'') the specified processes.
 
 %
 \declareAttribute{PMIX_JOB_CTRL_CANCEL}{"pmix.jctrl.cancel"}{char*}{
-Cancel the specified request (\code{NULL} implies cancel all requests from this requestor).
+Cancel the specified request - the provided request ID must match the \refattr{PMIX_JOB_CTRL_ID} provided to a previous call to \refapi{PMIx_Job_control}. An ID of \code{NULL} implies cancel all requests from this requestor.
 }
 
 %
--- a/Chap_API_Struct.tex
+++ b/Chap_API_Struct.tex
@@ -4526,7 +4526,7 @@ Attributes used to request control operations on an executing application - thes
 
 %
 \declareAttribute{PMIX_JOB_CTRL_ID}{"pmix.jctrl.id"}{char*}{
-Provide a string identifier for this request.
+Provide a string identifier for this request. The user can provide an identifier for the requested operation, thus allowing them to later request status of the operation or to terminate it. The host, therefore, shall track it with the request for future reference.
 }
 
 %
--- a/Chap_API_Struct.tex
+++ b/Chap_API_Struct.tex
@@ -3699,12 +3699,12 @@ This attribute would direct \refapi{PMIx_Finalize} to execute a barrier as part 
 
 %
 \declareAttribute{PMIX_JOB_TERM_STATUS}{"pmix.job.term.status"}{pmix_status_t}{
-Status to be returned upon job termination.
+Status returned by job upon its termination.
 }
 
 %
 \declareAttribute{PMIX_PROC_STATE_STATUS}{"pmix.proc.state"}{pmix_proc_state_t}{
-Process state
+State of the specified process as of the last report - may not be the actual current state based on update rate.
 }
 
 
--- a/Chap_Terms.tex
+++ b/Chap_Terms.tex
@@ -122,6 +122,12 @@ The call may update the argument but does not use its input value
 The call may both use and update the argument.
 \end{itemize}
 
+Many \ac{PMIx} interfaces, particularly nonblocking interfaces, use a \code{void*}cbdata object passed to the function that is then passed to the associated callback. In a client-side \ac{API}, the cbdata is a client-provided context (opaque object) that the client can pass to the nonblocking call (e.g., \refapi{PMIx_Get_nb}). When the nonblocking call (e.g., \refapi{pmix_value_cbfunc_t}) completes, the cbdata is passed back to the client without modification by the \ac{PMIx} library, thus allowing the client to associate a context with that callback. This is useful if there are many outstanding nonblocking calls.
+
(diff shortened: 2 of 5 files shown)


jjhursey:
Thanks for those clarifications. I'd like @dsolt 's 👍 before merging since he opened a couple of the clarifications.

Chap_API_Struct.tex
```diff
@@ -4526,7 +4526,7 @@ \subsection{Job control attributes}
 
 %
 \declareAttribute{PMIX_JOB_CTRL_ID}{"pmix.jctrl.id"}{char*}{
-Provide a string identifier for this request.
+Provide a string identifier for this request. The user can provide an identifier for the requested operation, thus allowing them to later request status of the operation or to terminate it. The host, therefore, shall track it with the request for future reference.
 }
```
dsolt: Should we add a note to PMIX_JOB_CTRL_CANCEL that specified request references a previous request marked with PMIX_JOB_CTRL_ID?

Chap_Terms.tex
```diff
@@ -122,6 +122,12 @@ \section{Procedure Conventions}
...
+
+A similar model is used for the server module functions (see \ref{server:module_fns}). In this case, the \ac{PMIx} library is making an upcall into its host via the \ac{PMIx} server module function and passing a specific cbfunc and cbdata. The \ac{PMIx} library expects the host to call the cbfunc with the necessary arguments and pass back the original cbdata upon completing the operation. This gives the server-side \ac{PMIx} library the ability to associate a context with the call back (since multiple operations may be outstanding). The host has no visibility into the contents of the cbdata object, nor is permitted to alter it in any way.
+
+\ac{PMIx} callback functions typically return \code{void} - i.e., they do not return a value and therefore cannot indicate the result of an operation. Errors encountered by \ac{PMIx} callback functions are to be communicated via the \refapi{PMIx_Notify_event} \ac{API} using a status code appropriate to the error. Implementers and users wishing to receive such notifications must register a corresponding event handler via the \refapi{PMIx_Register_event_handler} \ac{API}.
+
```
dsolt: Good clarifications on the first 2 paragraphs.   I don't quite get the value of the 3rd paragraph.  The server module functions all return a pmix_status_t which they can use to communicate failures.  They sometimes have a callback function, such as pmix_modex_cbfunc_t, which returns a void, but these functions are only used to communicate the results of the server operation.  If the callback function back into the PMIx server fails, this is presumably a serious defect/failure within the server as there is no situation where we would ever expect that to happen.   If we are taking about the client side, then again I don't understand how the client provided callbacks would ever need to raise an event or why that would even be the right course of action.   Maybe I need a real example of what you are thinking here?

Chap_API_Struct.tex
```diff
@@ -3699,12 +3699,12 @@ \subsection{Request-related attributes}
 
 %
 \declareAttribute{PMIX_JOB_TERM_STATUS}{"pmix.job.term.status"}{pmix_status_t}{
-Status to be returned upon job termination.
+Status returned by job upon its termination.
 }
```
dsolt: Do we document which API's might return PMIX_JOB_TERM_STATUS or PMIX_PROC_STATE_STATUS?   The descriptions are much improved, but in the version of the document I am looking at, I don't see when I would look for these?

Chap_Terms.tex
rhc54: I was responding to comments/questions from @rountree that were raised in the issue thread regarding "What should the error procedure be if cbdata is unrecognized?", which generalized into "what should the error procedure be if the callback function detects a problem?". Remember, the host does receive a number of callbacks from non-blocking function calls, and all of those callbacks return `void`. It could be a callback that returns the result of an operation, query of information or capabilities, etc.

The question shouldn't really be viewed as "how do I tell the PMIx library that something it gave me was wrong or indicated an error" - e.g., that a job control request failed. The question really is "how do I communicate that an error has occurred". Maybe I was looking for a specific capability and didn't find it, or I was trying to process the returned info and hit a data type that I don't recognize. Hence the use of `PMIx_Notify_event` - it lets both the target procs and the PMIx library know that something is wrong.

I can try to clarify it further based on the above explanation, if you find it helpful.

Chap_API_Struct.tex
rhc54: It's because they are returned in an event notification when the job or proc terminates, not from an API. I've been scratching my head a bit over this since you opened the cited issue as it isn't clear to me how to handle event-based attributes and values. They aren't really associated with any API. I was going to describe some of them in the tool chapter, but they really aren't tool-specific either - e.g., the ULFM folks use notifications of proc termination.

Perhaps the best approach is to expand the event notification chapter and cover them there? We could explain what attributes/values we expect the host to provide (e.g., job and proc termination events), how one requests that service, etc.

Would that make sense?

Chap_API_Struct.tex
rhc54: Yeah, that makes sense - will do!

rhc54: Done

Chap_Terms.tex
rhc54: I couldn't come up with a sensical way to explain it as your points were good, so I just removed the paragraph.

Chap_API_Struct.tex
rhc54: I described how these values are communicated in the attribute definition itself - see what you think.

jjhursey:
Per Sept. 5, 2019 teleconf - this is ok to merge when the author is ready unless @dsolt has any further comments.
@dsolt did you want to re-review the resolutions before the merge?

PR #218, opened 2019-09-06
Fix signature of blocking PMIx_Allocate_request

rhc54:
The blocking form of PMIx_Allocate_request was added in v3.1, but
unfortunately the signature didn't include the necessary parameters to
return the results of the request. This PR corrects that oversight.

Thanks to the Spark team for pointing it out.


jjhursey:
What do you think about putting this fix into v3? It's changing an API which seems a bit 'too big' for a minor version release, but that API that is there is not correct per this PR. I'm leaning towards bringing it into the v3.

We may want to sweep up other clarifications that have gone into v4 and not v3 at the same time and ready a point release for this fall.

rhc54:
Agreed - my thought was to release this in an errata v3.2 version. It is clearly an errata, and fortunately was caught before people had started to implement against it.

PR #222, opened 2019-09-14
Cleanup some blocking functions

rhc54:
Some of the blocking functions were missing necessary parameters to
return the results of their operation. In some casees, no blocking
form of an operation was provided - where possible, we provide a
blocking form by simply looking for a NULL callback function. We
return the result of the operation in the status code.

Need to bring at least some of these back to v3.2 errata

### DIFF
--- a/Chap_API_Security.tex
+++ b/Chap_API_Security.tex
@@ -41,7 +41,71 @@ Request a credential from the \ac{PMIx} server library or the host environment
 \begin{codepar}
 pmix_status_t
 PMIx_Get_credential(const pmix_info_t info[], size_t ninfo,
-                    pmix_credential_cbfunc_t cbfunc, void *cbdata)
+                    pmix_byte_object_t *credential)
+\end{codepar}
+\cspecificend
+
+\begin{arglist}
+\argin{info}{Array of \refstruct{pmix_info_t} structures (array of handles)}
+\argin{ninfo}{Number of elements in the \refarg{info} array (\code{size_t})}
+\argin{credential}{Address of a \refstruct{pmix_byte_object_t} within which to return credential (handle)}
+\end{arglist}
+
+Returns one of the following:
+
+\begin{itemize}
+    \item \refconst{PMIX_SUCCESS}, indicating that the credential has been returned in the provided \refstruct{pmix_byte_object_t}
+    \item a \ac{PMIx} error constant indicating either an error in the input or that the request is unsupported
+\end{itemize}
+
+\reqattrstart
+\ac{PMIx} libraries that choose not to support this operation \textit{must} return \refconst{PMIX_ERR_NOT_SUPPORTED} when the function is called.
+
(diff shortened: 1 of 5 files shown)


Chap_API_Security.tex
```diff
@@ -41,7 +41,71 @@ \subsection{\code{PMIx_Get_credential}}
 \begin{codepar}
 pmix_status_t
 PMIx_Get_credential(const pmix_info_t info[], size_t ninfo,
-                    pmix_credential_cbfunc_t cbfunc, void *cbdata)
+                    pmix_byte_object_t *credential)
```
jjhursey: Does this call need a set of output `info` and `ninfo` values (`pmix_info_t **results, size_t *nresults`) for the system returned set of values that would have been passed to the callback? From my reading, the `info`/`ninfo` passed into this function are not necessarily the same as those passed to the callback. If they are meant to be the same then it is fine as it is.

jjhursey:
Three of the functions impacted by this PR are nonblocking in nature but do not have the `_nb` suffix.
```
PMIx_Register_event_handler
PMIx_Deregister_event_handler
PMIx_Notify_event
```

This was because, at the time of introduction, we did not anticipate the need for a blocking form of the functions. However, since that time a corresponding blocking version has become needed to support Python. Since these functions are heavily used in their current form we are posed with a challenge of how to introduce a blocking version of the call.

After discussing this with Ralph, I agree that the path this PR is taking. To use a NULL reference passed to the callback as a signifier that this is a blocking operation. For these three functions, the return value is sufficient to provide the caller with the necessary information that they would have otherwise gotten from the callback. This is an easy way to add the blocking functionality without breaking existing usage of these APIs or introducing `_blocking` suffixed versions of the calls.

A similar adjustment was made to the following functions in this PR:
```
PMIx_IOF_pull
PMIx_IOF_deregister
PMIx_IOF_push
```

`PMIx_Job_control` and `PMIx_Process_monitor` had the same problem as was fixed for `PMIx_Allocation_request`, and are thus fixed in the same way (Ref PR #218).

The `PMIx_Get_credential` is being split into two functions. The prior API `PMIx_Get_credential` will become blocking (with a different argument set) and a new API `PMIx_Get_credential_nb` will be the nonblocking form (with the old signature). A similar action was taken with `PMIx_Validate_credential`. We need a note in the History about the signature change for these two items.

We should have a History update for all of these API modifications. Adding that as a separate commit to this PR will make it easier to cherry-pick just the changes to the v3 branch (since the v3 branch will need a custom patch for its history).

Chap_API_Security.tex
jjhursey: Per teleconf: The `credential` is the result so the `info` and `ninfo` are not needed.

rhc54:
I have rebased against master and provided the history update commit. Please note that ```PMIx_Get_credential``` and ```PMIx_Validate_credential``` were added in v3.0 of the Standard - we are simply correcting the signature here.

PR #224, opened 2019-09-26
Add a little more info and examples to Python bindings

rhc54:
Extend the Python binding appendix with a few notes on design
details and a couple of examples.


dsikich:
@rhc54 @jjhursey I think that the group server module function may be missing from this draft in the python bindings section.

rhc54:
@dsikich Nice catch! Added commit to fill the gap.

PR #226, opened 2019-10-15
Fix two identical typos

gvallee:
"rwere" -> "were"


rhc54:
Just curious - why close this?

PR #229, opened 2019-10-29
Add a watermark for Unofficial Draft builds.

### LINKED ISSUE #228: Add a watermark for draft copies of the standard
We often have draft copies of the standard floating around as we evaluate the new text. It is helpful to have a watermark of some sort to indicate that the PDF is an unofficial draft so users that stumble on the PDF know that it is not official.

jjhursey:
* Fixes #228

### DIFF
--- a/pmix-standard.tex
+++ b/pmix-standard.tex
@@ -44,14 +44,37 @@
 % The following says letter size, but the style sheet may change the size
 \documentclass[10pt,letterpaper,twoside,makeidx,hidelinks]{scrreprt}
 
+%%%%%%%%%%%%%%%%%%%
+% Release Managers:
+% - Set is_unofficial_draft = false to remove the watermark
+% - Set VER to the correct version number
+% - Set VERDATE to the Month+Year of the release
+%%%%%%%%%%%%%%%%%%%
+\usepackage[us,24hr]{datetime}
+\usepackage{ifthen}
+\newboolean{is_unofficial_draft}
+\setboolean{is_unofficial_draft}{true}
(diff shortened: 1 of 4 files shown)


jjhursey:
I don't love that I had to include `bookmark.sty` but I couldn't find my way around the `RequirePackage{hyperref}` issue.

pmix-standard.tex
```diff
@@ -44,15 +44,38 @@
...
+%%%%%%%%%%%%%%%%%%%
+% Release Managers:
+% - Set is_unofficial_draft = false to remove the watermark
+% - Set VER to the correct version number
+% - Set VERDATE to the Month+Year of the release
+%%%%%%%%%%%%%%%%%%%
+\usepackage[us,24hr]{datetime}
+\usepackage{ifthen}
+\newboolean{is_unofficial_draft}
+\setboolean{is_unofficial_draft}{true}
```
SteVwonder: What is the intended procedure for unsetting/setting this boolean during a release?  Merge a PR that sets it to `false`, tag the release, then shortly after merge another PR that sets it back to `true`?

rhc54: Generally, yes - you would set it back to true _and_ up the revision number (adding an "a1" or something to indicate it is a working branch again) in a subsequent PR.

jjhursey: I'd like the default value in the repo to be `true` so that if someone checks out the repo they will get the watermark. For the release tag point having it set to `false` in one commit seems reasonable to me.

Issue #175, opened 2019-03-29
Further separate out the PMIx client interface from the CL/host interactions


dsolt:
The PMIx standard defines a set of PMIx API's that are used to allow applications (clients) to communicate with a runtime environment.  However, interwoven within the presentation of that API is the concept of three distinct components:  a PMIx host server, a PMIx client library and a host RM (resource manager).  The distinction between the client library and a host server is  important because it would be of great benefit if an application could be linked against one implementation's PMIx client library and run on another system with a different implementation of the server.   However, the implementation of the host server and the distinction between a host PMIx server and a host RM component should have no impact on application developers (if it does, it leads to non-portable concepts that benefit only one implementation of PMIx).   Although the standard nicely attempts to separate out the API between the host server and host RM (See chapter 11 "Server-specific interfaces"), this architectural assumption can still be found spread through the document.   This issue is to explore the idea of further separating out these concepts, which would lead to at least the following benefits:

1) A application developer can more easily navigate the standard to understand how to use the PMIx APIs
2) Developers implementing the standard who wish to use the architecture of a separate host server + RM can easily find and understand what is required to implement that host PMIx server and understand what is required to implement the RM using a separate host PMIx server.
3) Developers implementing the standard who wish to directly implement the PMIx API can understand what is required without having to separate out text that is specific to the host server/host RM architecture.  
4) The separation will discourage concepts from being part of the PMIx API that are specific to  a host server/host RM architecture. 

To make this less nebulous, take a moment to look at a concrete example.

Standard 3.1, sections 3.4.17, 3.4.18 - These sections describe concepts that are completely irrelevant to an application developer (and may be very confusing for someone reading the document from that point of view), but they contain information that may be very important and should not simply be discarded.  The question I think we should ask is, 1) Is this information important between client and server.  If so, we should move it to a section that is clearly marked as important only for an implementer of the client or server.  If the information is important only host PMIx server and host RM, then that information should be moved to a section and clearly marked as important only for an implementer using the host server/host  RM architecture.   

I propose a thorough scouring of the document by a working group to look for all possible areas that could be sectioned off as relevant to  1)  client/server interactions 2) host PMIx server/host RM architecture implementations.

rhc54:
[...]

PR #192, opened 2019-06-04
Ch. 1: Issue 175:  Separate out the PMIx client interface from the CL/host interactions 

dsolt:
…y searching

PR #230, opened 2019-12-04
Ch. 4->3: Clarify three roles of the interface

jjhursey:
* Effort to clarify the three roles of consumers of the PMIx interface (clients / servers / tools)
 * Goal is to use the new Chapter 3 as a splitting point to direct those focused on a specific role to those sections of the standard that are most meaningful to them.
 * Part of the implementation agnostic/client separation working group effort
 * Ref Issue #175

### DIFF
--- a/Chap_API_Init.tex
+++ b/Chap_API_Init.tex
@@ -4,27 +4,44 @@
 \chapter{Initialization and Finalization}
 \label{chap:api_init}
 
-The \ac{PMIx} library is required to be initialized and finalized around the usage of most of the \acp{API}.
-The \acp{API} that may be used outside of the initialized and finalized region are noted.
-All other \acp{API} must be used inside this region.
+The \ac{PMIx} \acp{API} may only be used between the completion of the initialization function and the start of the finalization function, unless otherwise noted.
+The initialization and finalization functions are paired, and the initialized regions defined by them must not overlap.
 
-There are three sets of initialization and finalization functions depending upon the role of the process in the \ac{PMIx} universe.
-Each of these functional sets are described in this chapter.
-Note that a process can only call \textit{one} of the init/finalize functional pairs - e.g., a process that calls the client initialization function cannot also call the tool or server initialization functions, and must call the corresponding client finalize.
+Consumers of the \ac{PMIx} \acp{API} are grouped into three categories based on the role they wish to play in the \ac{PMIx} environment: \emph{clients}, \emph{servers}, and \emph{tools}.
+As a result, there are three corresponding sets of initialization and finalization functions.
+If a process initializes as either a \emph{server} or a \emph{tool} that process may also access all of the \emph{client} \acp{API}.
 
-\adviceuserstart
-Processes that initialize as a server or tool automatically are given access to all client \acp{API}.
-Server initialization includes setting up the infrastructure to support local clients - thus, it necessarily includes overhead and an increased memory footprint.
-Tool initialization automatically searches for a server to which it can connect --- if declared as a \textit{launcher}, the \ac{PMIx} library sets up the required ``hooks'' for other tools (e.g., debuggers) to attach to it.
-\adviceuserend
+A process operating as a \emph{client} is started (directly or indirectly, for example, by an intermediate script) by a \ac{PMIx} server and is connected to that server when the client calls the client initialization routine.
+A process operating as a \emph{server} is responsible for starting client processes and coordinating with other server and tool processes in the same \ac{PMIx} universe.
+Often processes operating as a \emph{server} are part of the \acf{RM} infrastructure.
+A process operating as a \emph{tool} will connect to a \ac{PMIx} \emph{server} to interact with the processes in the \ac{PMIx} universe.
+An example of a \emph{tool} process is a parallel debugger that will connect to the server to assist with attaching to a set of client processes.
+
+\ac{PMIx} serves as a conduit between processes acting in these three different roles.
+As such, an \ac{API} is often described in how it interacts with processes operating in other roles in the \ac{PMIx} universe.
+
+\adviceimplstart
+A \ac{PMIx} implementation may support all or a subset of the \ac{API} role groupings defined in the standard.
+A common nomenclature is defined here to aid in identifying levels of conformance of an implementation.
+
+A \ac{PMIx} implementation that supports all three sets of the \ac{API} role groupings is said to be \emph{fully \ac{PMIx} standard compliant}.
+These \emph{fully \ac{PMIx} standard compliant} implementations have the advantage of being able to support a broad set of \ac{PMIx} consumers in the different roles.
+
+Alternatively, a \ac{PMIx} implementation may choose to support fewer than all three sets of the \ac{API} role groupings.
+\ac{PMIx} implementations that support only the \emph{client} \acp{API} are said to be \emph{client-only \ac{PMIx} standard compliant}.
+Similarly, an implementation that only supports the \emph{client} and \emph{tool} \acp{API} are said to be \emph{client-and-tool-only \ac{PMIx} standard compliant}.
+Finally, an implementation that only supports the \emph{client} and \emph{server} \acp{API} are said to be \emph{client-and-server-only \ac{PMIx} standard compliant}.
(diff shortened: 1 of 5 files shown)


jjhursey:
I think this is ready for the Working Group to start reviewing.

jjhursey:
Here is a link to the generated PDF that I will try to keep updated with any further changes:
 * https://ibm.box.com/s/i1sar8jl62euzbj29ybj9bhp5zae0yqv

Chap_API_Init.tex
```diff
@@ -4,27 +4,44 @@
...
-Note that a process can only call \textit{one} of the init/finalize functional pairs - e.g., a process that calls the client initialization function cannot also call the tool or server initialization functions, and must call the corresponding client finalize.
+Consumers of the \ac{PMIx} \acp{API} are grouped into three categories based on the role they wish to play in the \ac{PMIx} environment: \emph{clients}, \emph{servers}, and \emph{tools}.
+As a result, there are three corresponding sets of initialization and finalization functions.
+If a process initializes as either a \emph{server} or a \emph{tool} that process may also access all of the \emph{client} \acp{API}.
+
+A process operating as a \emph{client} is started (directly or indirectly, for example, by an intermediate script) by a \ac{PMIx} server and is connected to that server when the client calls the client initialization routine.
+A process operating as a \emph{server} is responsible for starting client processes and coordinating with other server and tool processes in the same \ac{PMIx} universe.
+Often processes operating as a \emph{server} are part of the \acf{RM} infrastructure.
```
dsolt: should we define PMIx universe in ch2?

Chap_API_Init.tex
```diff
@@ -4,27 +4,44 @@
...
 
-There are three sets of initialization and finalization functions depending upon the role of the process in the \ac{PMIx} universe.
-Each of these functional sets are described in this chapter.
-Note that a process can only call \textit{one} of the init/finalize functional pairs - e.g., a process that calls the client initialization function cannot also call the tool or server initialization functions, and must call the corresponding client finalize.
+Consumers of the \ac{PMIx} \acp{API} are grouped into three categories based on the role they wish to play in the \ac{PMIx} environment: \emph{clients}, \emph{servers}, and \emph{tools}.
+As a result, there are three corresponding sets of initialization and finalization functions.
+If a process initializes as either a \emph{server} or a \emph{tool} that process may also access all of the \emph{client} \acp{API}.
+
+A process operating as a \emph{client} is started (directly or indirectly, for example, by an intermediate script) by a \ac{PMIx} server and is connected to that server when the client calls the client initialization routine.
```
rhc54: Bit of a nit here, but the PMIx server doesn't actually "start" client processes - the host is responsible for doing that.

jjhursey: Good point. Would it be more accurate to say:
 > A process operating as a \emph{client} is started (directly or indirectly, for example, by an intermediate script) by the \ac{RM} and is connected to the \ac{PMIx} server instance within that \ac{RM} when the client calls the client initialization routine.

Chap_API_Init.tex
jjhursey: Yeah, we probably should. It seems to be used in the document to refer to the set of all PMIx entities which interact together possibly only within a session - but I don't know if it is exclusive to the session. If it is exclusive then we could just say `the same PMIx session`.

Chap_API_Init.tex
rhc54: Yeah, that works - sorry for the extra verbosity, but it seems to be a regular source of confusion so merits extra care.

Chap_API_Init.tex
rhc54: The "universe" was meant to be the collection of sessions under the control of a common scheduler or SMS - usually that will mean "everything on the cluster", but could be broader when you extend to the grid environment. IIRC, there was a sentence or two somewhere that tried to capture that definition (perhaps in one of the papers, if not in this doc?).

jjhursey: I added a comment to the Ch2 PR so we capture this definition.
 * https://github.com/pmix/pmix-standard/pull/235#issuecomment-584671141

Chap_API_Init.tex
jjhursey: Change made. Thanks!

jjhursey:
Per Client Separation Working Group Meeting on June 9, 2020:
 * Move the server init/finalize routines to the "Server-specific interfaces" chapter
 * Move the 'new' preamble text to Chapter 1. This defines the three roles in the interface.
   - Note once PR #192 is merged (next month) then I'll replace this section within the new Chapter 1 text.
 * Move the general information interfaces to a new chapter out of the init/finalize chapter
 * Make the init/finalize chapter a "Client-Specific Interfaces" chapter
   - For now it includes the tool init/finalize routines. But once we get the Tools chapter then we will move those interfaces to the new chapter.
 * Idea is that we will have chapters dedicated to interfaces specific to each of the three roles. Then the rest of the chapters are generally consumable (more or less) by any role. it also provides a specific chapter for folks to jump to if they are only concerned with one of the three roles.

jjhursey:
Per WG teleconf June 15, 2020
 * Section 1.3.3 (new section) - General note that 'consumer' is misleading in the title, at least in the current form.
   - Change the title to (here are some options - vote with emoji)
      - 👍  "Types of PMIx API Consumers"
      - 😄  "Categories of PMIx Roles"
      - 🎉  "PMIx Roles"
      - 🚀  "PMIx API Role Groupings"
      - 👎  -- back to the drawing board - suggest in comments or in the next meeting
   - Change the first sentence to
     - "The role of a PMIx process in the PMIx universe is grouped into three categories based on how it operates in the PMIx environment namely as a client, server, or tool"
 * New Chapter 3 ("General Information Interfaces) is a bit small
   - Keep it for now to see if other things fit into this category (maybe renaming the chapter)
   - Maybe PMIx_Query should move in here?
 * New Chapter 4 ("Client-Specific Interfaces") is also small since there are not many interfaces exclusive to clients that cannot also be used by tools and servers.
   - That's ok for now, but seems a bit odd.
 * All other chapters are generally "part of the client role" of the interface - meaning that tools and servers can also use them. Should we put a note in each chapter designating the role in which it is intended to be used?

The current version of the document for reference is attached to this comment.
[pmix-standard.pdf](https://github.com/pmix/pmix-standard/files/4782563/pmix-standard.pdf)

jjhursey:
I updated the title and first sentence of Section 1.3.3 - attached in the current rendering.
[pmix-standard.pdf](https://github.com/pmix/pmix-standard/files/4814804/pmix-standard.pdf)

jjhursey:
Per teleconf today - Slight adjustment to first sentence wording.
[pmix-standard.pdf](https://github.com/pmix/pmix-standard/files/4814891/pmix-standard.pdf)

jjhursey:
We might bring in part of this PR into the PMIx v4.0 standard. We can discuss in the next monthly ASC meeting.

Chap_API_Init.tex
```diff
@@ -372,8 +311,12 @@ \subsection{\code{PMIx_tool_connect_to_server}}
 %%%%
 \descr
 
-Switch connection from the current \ac{PMIx} server to another one, or initialize a connection to a specified server. Closes the connection, if existing, to a server and establishes a connection to the specified server. This function can be called at any time by a \ac{PMIx} tool to shift connections between servers.
-The process identifier assigned to this tool is returned in the provided \refstruct{pmix_proc_t} struct. Passing a value of \code{NULL} for this parameter is allowed if the user wishes solely to connect to the \ac{PMIx} server and does not require return of the identifier at that time.
+A tool may call \refapi{PMIx_tool_init} with the \refattr{PMIX_TOOL_DO_NOT_CONNECT} attribute in which case they can use this function to connect to a specific server.
+Additionally, a tool may use this function to switch connection from the current \ac{PMIx} server to another one
```
naughtont3, line 315: Seems like something is missing at end of this sentence, leading to next line.

jjhursey: Humm. Yeah, I think I missed something here. I'll fix and push in a minute. Thanks!

jjhursey: Fixed in latest push

jjhursey:
This PR has grown stale. It tried to do 2 things:
 * Split up the initialization/finalization of client / tool / server into separate chapters. This was completed in v4.
 * Define the 3 consumers of the interface. I pulled this out into PR #327

PR #241, opened 2020-06-09
Split index into multiple parts and add automated document check

### LINKED ISSUE #149: Automated document checker
This request is being filed in both the PMIx Standards and PMIx Reference Implementation repositories, with cross-references between them.

We need a Python script that can cross-check the Standards document with the corresponding PRI version to detect attributes, constants, macros, and functions that are either:

- defined in the Standard but not present in the PRI
- present in the PRI but not contained in the Standard

This is becoming critical as the document grows in both size and the number of versions.

jjhursey:
* Multiple indices will make it easier to find APIs vs attributes vs ...
 * Add an automated document cross-check against OpenPMIx
   - Checks PMIx Standard items not included in OpenPMIx
   - Checks OpenPMIx items not included in PMIx Standard

Closes #149

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 *.xwm
 datetime-defaults.sty
 datetime.sty
+check-openpmix
--- a/App_Python.tex
+++ b/App_Python.tex
@@ -1008,7 +1008,7 @@ The client Python class is by far the richest in terms of \acp{API} as it houses
 
 %%%%%%%%%%%
 \subsection{Client.init}
-\declareapi{PMIxClient.init}
+\declareapibinding{PMIxClient.init}{PMIx_Init}{Python}
 
 \summary Initialize the \ac{PMIx} client library after obtaining a new PMIxClient object
 
@@ -1038,7 +1038,7 @@ See \refapi{PMIx_Init} for description of all relevant attributes and behaviors
 
 %%%%%%%%%%%
 \subsection{Client.initialized}
-\declareapi{PMIxClient.initialized}
+\declareapibinding{PMIxClient.initialized}{PMIx_Initialized}{Python}
 
 \format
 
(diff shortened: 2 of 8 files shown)


jjhursey:
Ref Issue #149 - this will generate .idx files that can be used to access the attributes/apis/macros/constants/structs from the standard. That will be useful when building the document checker.

Keeping this a WIP for now as I work on the document checker script.

jjhursey:
[pmix-standard-pr-241-v1.pdf](https://github.com/pmix/pmix-standard/files/4754889/pmix-standard-pr-241-v1.pdf)
Current rendering of this PR if folks want to check it out.

jjhursey:
PDF rendering of the current state of the PR (so you can see the indexing)
[pmix-standard-pr-241-v2.pdf](https://github.com/pmix/pmix-standard/files/4766421/pmix-standard-pr-241-v2.pdf)

jjhursey:
I plan to discuss the indexing changes in the PMIx Standard meeting on Friday, June 11. So I won't plan to merge before then.

This is missing one optional feature of (we don't need it right now so we can wait on it)
   - Whitelisting of items that are ok to be missing on either side (e.g., something in OpenPMIx that is not meant to be part of the PMIx Standard or vise versa)

SteVwonder:
Very nice @jjhursey! I wonder if we can leverage this work to improve the [coverage analysis](https://github.com/SteVwonder/functionality-WG/blob/master/src/coverage.org) and cleanup some of the inconsistencies that @rhc54 pointed out.

jjhursey:
Per teleconf June 12, 2020
 * Can we remove the `Definition` line, and instead have a bolded page number at the front of the reference list? That would compress the index a bit, and still preserve meaning.
 * Generally like the breakdown in the current index. Generally ok with the mix of double and single-column format.

jjhursey:
[pmix-standard.pdf](https://github.com/pmix/pmix-standard/files/4782965/pmix-standard.pdf)

This version uses bold and underlining to denote the Definition in the index (instead of a separate Definition line). I moved all the indices to single column as that helped scanning down the list for a given item.

I think this is ready for review for inclusion in v4.

PR #243, opened 2020-06-12
Add/update the fabric support chapter

rhc54:
Consolidate the scheduler and coordinate chapters into a single chapter
on fabric support. Pull all fabric-related APIs, data types, constants,
and attributes into the new chapter. Update the definition of the
pmix_fabric_t object to be a little more abstract. Provide a new API for
requesting an update to the fabric object for systems that need to poll
for updates rather than use the event notification method.


jjhursey:
Feel free to merge when ready.

PR #244, opened 2020-06-17
Add a server module function for fabric requests

rhc54:
Fabric requests can be made by tools as well as servers, so we need an
upcall to allow the host system to route the request to the daemon
hosting that support.

### DIFF
--- a/Chap_API_Server.tex
+++ b/Chap_API_Server.tex
@@ -1111,6 +1111,7 @@ typedef struct pmix_server_module_3_0_0_t {
     pmix_server_stdin_fn_t              push_stdin;
     /* v4x interfaces */
     pmix_server_grp_fn_t                group;
+    pmix_server_fabric_fn_t             fabric;
 } pmix_server_module_t;
 \end{codepar}
 \cspecificend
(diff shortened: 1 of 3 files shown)


Chap_API_Server.tex
```diff
@@ -2629,10 +2630,6 @@ \subsection{\code{pmix_server_monitor_fn_t}}
 
 Request that a client be monitored for activity.
 
-\advicermstart
-If this module entry is provided and called by the \ac{PMIx} server library, then the host environment must either provide the requested services or return \refconst{PMIX_ERR_NOT_SUPPORTED} to the provided \refarg{cbfunc}.
-\advicermend
```
jjhursey, line 2634: I'm curious why you removed this advice to RMs. It seems redundant to have it for some of the module functions (makes them seem more optional than others).

I'm fine with removing it from the individual module functions, but maybe we should put a statement near the `pmix_server_module_t` structure generally stating this. What do you think?

rhc54: It bothered me because we specify the return values up above that explains what this function should return - we can callout "not supported" there if we think it needs to be done. I think there is a general statement about what happens if the entry isn't there - we can add a clause to that if we want.

rhc54: Just pushed a modification to address this - see what you think.

jjhursey: That looks good to me

PR #245, opened 2020-06-17
Fix check scripts

jjhursey:
* Fix check-ref-attr now that we no longer use "Definition" in the index
 * For deprecated attributes add a "(Deprecated)" note to the index
 * Fix check-openpmix to better match names when double checking to see if something is missing


jjhursey:
Here is an image showing some deprecated attributes as they would appear in the index:

image

PR #249, opened 2020-06-25
Make check-openpmix more robust

jjhursey:
* Dynamically extact the header files (so if they change in the
   future we don't have to adjust this script)
 * If the check-openpmix checkout already exists then make sure to do
   a 'git pull' on it before checking.
 * Make a fake target so the check targets always rebuild


jjhursey:
No document changes, just infrastructure in this PR.

Issue #242, opened 2020-06-11
Missing Items from 'check-openpmix'


jjhursey:
In PR #241 (which fixes Issue #149) I identified a number of items missing both from OpenPMix and the PMIx Standard. Below are the current lists against PMIx Standard `master` and OpenPMIx `master`

Use `make check-openpmix` to run this test.

<details>
 <summary>(98) Items in OpenPMIx that are missing from the PMIx Standard</summary>

```
--------------------------------------------------
Found 98 references defined in OpenPMIx, but not in the PMIx Standard
PMIx Standard Missing: PMIX_ALLOC_QUEUE
PMIx Standard Missing: PMIX_APP_ARGV
PMIx Standard Missing: PMIX_APP_RELEASE
PMIx Standard Missing: PMIX_ARGV_PREPEND
PMIx Standard Missing: PMIX_BYTE_OBJECT_CONSTRUCT
PMIx Standard Missing: PMIX_CMD_LINE
PMIx Standard Missing: PMIX_DATA_TYPE_MAX
PMIx Standard Missing: PMIX_DEBUG_TARGET
PMIx Standard Missing: PMIX_ERR_BASE
PMIx Standard Missing: PMIX_ERR_CONFLICTING_CLEANUP_DIRECTIVES
PMIx Standard Missing: PMIX_ERR_JOB_ABORTED
PMIx Standard Missing: PMIX_ERR_JOB_ABORTED_BY_SIG
PMIx Standard Missing: PMIX_ERR_JOB_ALLOC_FAILED
PMIx Standard Missing: PMIX_ERR_JOB_APP_NOT_EXECUTABLE
PMIx Standard Missing: PMIX_ERR_JOB_CANCELLED
PMIx Standard Missing: PMIX_ERR_JOB_CANNOT_LAUNCH
PMIx Standard Missing: PMIX_ERR_JOB_FAILED_TO_LAUNCH
PMIx Standard Missing: PMIX_ERR_JOB_FAILED_TO_MAP
PMIx Standard Missing: PMIX_ERR_JOB_FAILED_TO_START
PMIx Standard Missing: PMIX_ERR_JOB_KILLED_BY_CMD
PMIx Standard Missing: PMIX_ERR_JOB_NEVER_LAUNCHED
PMIx Standard Missing: PMIX_ERR_JOB_NON_ZERO_TERM
PMIx Standard Missing: PMIX_ERR_JOB_NO_EXE_SPECIFIED
[...]

Issue #231, opened 2019-12-18
Reference to pmix.org website in `PMIx_Get` Description


SteVwonder:
In the "Advice to Users" section of the `PMIx_Get` description, there is the following sentence:
> The list of data referenced in this way is maintained on the PMIx web site at https://pmix.org/support/faq/wildcard-rank-access/ but includes items such as the number of processes in the namespace ( PMIX_JOB_SIZE ), total available slots in the allocation ( PMIX_UNIV_SIZE ), and the number of nodes in the allocation (PMIX_NUM_NODES ).

When you visit that link, it contains only this sentence:
> Information accessed via calls to PMIx_Get or PMIx_Get_nb require specification of both the nspace and rank of the publishing process.

I do not think the content on the current version of that webpage adds much to the description of the interface. In addition, this redirect to the website prevents the standard from being self-contained.

### Suggested Clarification

I am suggesting to remove the url from the standards document and instead enumerate the pieces of data accessed in this way directly within the standard.  Unless of course that is an implementation detail; if so, we should brainstorm on the best way to handle that.

### References

Chapter 5, Page 129, Line 17 of the current master (ba66b70f038fc8a08d36c009bec0caa3b2d16516)

jjhursey:
This was fixed in PR #253 and is included in the upcoming v5.0 release and in v4.1 release.

PR #253, opened 2020-07-04
Add new chapter on tools and perform cleanup

rhc54:
Add an early scaffold of the new chapter on tools - this should be ignored for now.

Cleanup a variety of things:
* add missing attributes found with check-openpmix.py (Refs #242)
* add missing job-related state and error constants found with check-openpmix.py (Refs #242)
* cleanup registration info description (Refs #231)
* remove all references to PRI and PRRTE attribute/constant designations from some of the chapters (not a full culling of the Standard)
* add a few clarifications to PMIx_Get


rhc54:
@SteVwonder Please see if this adequately addressed the concerns of #231

Chap_API_Key_Value_Mgmt.tex
```diff
@@ -128,19 +128,12 @@ \subsection{\code{PMIx_Get}}
 %%%%
 \descr
 
-Retrieve information for the specified \refarg{key} as published by the process identified in the given \refstruct{pmix_proc_t}, returning a pointer to the value in the given address.
+Retrieve information for the specified \refarg{key} associated with the process identified in the given \refstruct{pmix_proc_t}, returning a pointer to the value in the given address. In general, data posted by a process via \refapi{PMIx_Put} and data that refers directly to a process-related value must be retrieved by specifying the rank of the target process. All other information is retrievable using a rank of \refconst{PMIX_RANK_WILDCARD}, as illustrated in \ref{chap:api_kv:getex}. See \ref{api:struct:attributes:retrieval} for an explanation regarding use of the \emph{level} attributes, and \refapi{PMIx_server_register_nspace} for a description of the available information.
```
SteVwonder, line 131: For any "pre-populated" information that wasn't necessarily `PMIx_Put` by another process/rank, is there a section within the standard that lists the reserved keys for that info?  Is Section 3.4 (Reserved Attributes) that section (maybe 3.47 - 3.4.14 more specifically)?  If so, it might be good to cross-reference that.  After reading through just the PMIx_Get description, it isn't clear to me necessarily where I should go to learn more about what pre-populated information I can get via this interface (subject to implementation support of course).  And I think the  `api:struct:attributes:retrieval` section is just a subset of the possible pre-populated information, correct?

rhc54: By definition, info "put" by another process/rank can't be in the pre-populated info and therefore any such info cannot use a reserved key - I'll try to make that clearer and reference Section 3.4.

IIRC, I was intending `api:struct:attributes:retrieval` section to contain all _required_ pre-populated info - I'll double check and ensure it does, adding anything that is missing. I'll also be sure to add all the defined _optional_ info keys so it provides a complete picture.

PR #256, opened 2020-07-27
Update v4.0 history and add process locality support

rhc54:
Update v4.0 History to include latest changes

Add process locality definitions and APIs

### DIFF
--- a/Chap_API_Fabric.tex
+++ b/Chap_API_Fabric.tex
@@ -70,7 +70,7 @@ Nodes with multiple fabric devices can also have those devices configured as mul
 The following macros are provided to support the \refstruct{pmix_coord_t} structure.
 
 %%%%
-\subsubsection{Initialize the \refstruct{pmix_coord_t} structure}
+\subsubsection{Initialize the coord structure}
 \declaremacro{PMIX_COORD_CONSTRUCT}
 
 Initialize the \refstruct{pmix_coord_t} fields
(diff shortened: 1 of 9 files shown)


rhc54:
@jjhursey Can you please take a look? I believe the locality definitions are consistent with what we discussed.

rhc54:
@jjhursey OMPI currently doesn't need to convert the locality string back to a cpuset - we only use relative locality to guide shared memory setup. The problem with converting it back is that it would require access to the full topology tree, which is what the locality string is trying to prevent (otherwise, we would just pass down the cpuset itself and let the client use the topology tree to determine locality).

We do make the topology tree potentially available via the shmem and xml representation attributes, so someone could always go that way instead. I can add a few sentences explaining why we offer locality and the alternatives, if you think that would be helpful.

rhc54:
@jjhursey Sorry about all the whitespace removal - my editor automatically filters those out and cleans up any file I touch. Both good and bad, I guess 🤷‍♂️

rhc54:
I incorporated your comments, added a little more clarification about the meaning of locality, and did a touch of reorganization. I also extended the TOC to include subsubsections to make it easier to find things, and made the "typeDEP" and "typeNEW" suffixes uniform as I had to constantly refer back to pmix.sty to remember which type used what form.

rhc54:
@jjhursey I reorganized and did a little more cleanup - see what you think. I attached just the relevant PDF section here.

[locality.pdf](https://github.com/pmix/pmix-standard/files/4988747/locality.pdf)
[locserver.pdf](https://github.com/pmix/pmix-standard/files/4988749/locserver.pdf)

jjhursey:
I think this looks good. Thanks for adding the clarifications.

I see your point about converting the string back to a cpuset. Since we don't have a use case driver for it we can skip it. If there is a driver in the future then they need to think about the implications of loading the topology in their proposal. I don't have a need for it in mind.

PR #257, opened 2020-07-29
Extend topology support

rhc54:
Users and applications would like a scalable way to retrieve the local
topology, preferably via a shared memory region from the local
persistent daemon. This alleviates the discovery time problem for large
ppn scenarios as well as the memory footprint penalty associated with
every proc carrying its own topology description.

Define a new pmix_topology_t abstraction that includes the source of the
topology description (e.g., "hwloc") and a void* pointer to the
description itself. Add a PMIX_TOPOLOGY2 attribute to use that value
in place of hwloc_topology_t to provide more of an abstraction.
Deprecate PMIX_TOPOLOGY.  Server library is responsible for cleaning up
at finalize.
    
When PMIX_TOPOLOGY or PMIX_TOPOLOGY2 is passed to a server during
PMIx_server_init, have the server be responsible for setting up whatever
support is available (depending on topology implementation) and passing
the required keys down to the client. This removes all the current PMIX_*
topology keys as those become hidden inside the PMIx implementation -
so deprecate them in this version.

Add a new PMIx_Load_topology API by which the user/app can retrieve the
topology description for their subsequent use. Strongly warn that the
returned pointer is to be treated as "read only" and that attempts to
modify it will return an error. Client library is responsible for
cleaning up at finalize.

Do a little cleanup of long lines so diff is easier to read. Fix an
error in naming convention - it should be PMIx_Get_relative_locality
with a capital-G as it is a client-facing function.

### DIFF
--- a/Chap_API_Proc_Mgmt.tex
+++ b/Chap_API_Proc_Mgmt.tex
@@ -607,10 +607,67 @@ Nonblocking \refapi{PMIx_Disconnect} routine. The callback function is called on
 
 The relative locality of processes is often used to optimize their interactions with the hardware and other processes. \ac{PMIx} provides a means by which the host environment can communicate the locality of a given process using the \refapi{PMIx_generate_locality_string} to generate an abstracted representation of that value. This provides a human-readable format and allows the client to parse the locality string with a method of its choice that may differ from the one used by the server that generated it.
 
+There are times, however, when relative locality and other \ac{PMIx}-provided
+information doesn't include some element required by the application. In these
+instances, the application may need access to the full description of the
+local hardware topology. \ac{PMIx} does not itself generate such descriptions
+- there are multiple third-party libraries that fulfill that role. Instead,
+\ac{PMIx} offers an abstraction method by which users can obtain a pointer to
+the description. This transparently enables support for different methods of
+sharing the topology between the host environment (which may well have already
+generated it prior to local start of application processes) and the clients -
+e.g., through passing of a shared memory region.
 
 %%%%%%%%%%%
-\subsection{\code{PMIx_get_relative_locality}}
-\declareapi{PMIx_get_relative_locality}
+\subsection{\code{PMIx_Load_topology}}
+\declareapi{PMIx_Load_topology}
+
+%%%%
+\summary
+
+Load the local hardware topology description
+
+%%%%
+\format
+
+\versionMarker{4.0}
+\cspecificstart
+\begin{codepar}
+pmix_status_t
+PMIx_Load_topology(pmix_topology_t *topo);
+\end{codepar}
+\cspecificend
+
+\begin{arglist}
+\arginout{topo}{Address of a \refstruct{pmix_topology_t} structure where the topology information is to be loaded (handle)}
+\end{arglist}
+
+Returns \refconst{PMIX_SUCCESS}, indicating that the \refarg{topo} was successfully loaded, or an appropriate \ac{PMIx} error constant.
+
+%%%%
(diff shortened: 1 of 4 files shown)


Chap_API_Server.tex
```diff
@@ -80,6 +96,13 @@ \subsection{\code{PMIx_server_init}}
 \pasteAttributeItemBegin{PMIX_SERVER_REMOTE_CONNECTIONS} If the library supports connections from remote tools, this attribute may be supported for enabling or disabling it.
 \pasteAttributeItemEnd{}
 \pasteAttributeItem{PMIX_EVENT_BASE}
+\pasteAttributeItemBegin{PMIX_TOPOLOGY}If provided, the \ac{PMIx} server will
```
jjhursey, line 99: Should this be `PMIX_TOPOLOGY2`?

Chap_API_Proc_Mgmt.tex
```diff
@@ -607,10 +607,69 @@ \section{Process Locality}
...
+the \ac{PMIx} library must return a description from the specified
+implementation or else indicate that the implementation is not available by
+returning the \refconst{PMIX_ERR_NOT_SUPPORTED} error constant.
+
+The returned pointer may point to a shared memory region or an actual instance
+of the topology description. In either case, the description shall be treated
+as a "read-only" object - attempts to modify the object are likely to fail and
+return an error. The \ac{PMIx} library is responsible for reference counting
+the returned object and performing any required cleanup either when the
+reference count reaches zero or the library itself finalizes.
```
jjhursey, line 661: How would a PMIx library go about reference counting the returned object? We could increment the reference count when giving it to the client, but we don't have a 'release' function for the client to indicate that they are done with it. Maybe we just need a `PMIX_TOPO_RELEASE` call - and a sentence here like "when the returned object is no longer required then the caller must pass it to `PMIX_TOPO_RELEASE` to ensure safe cleanup."

Chap_API_Server.tex
rhc54: Yes - thx!

Chap_API_Proc_Mgmt.tex
rhc54: In thinking about it, we don't need this at all. We can just release the topology during finalize since that is already protected by internal refcount. We can internally track the source of the topology (i.e., whether it came from shmem/xml or was independently discovered) and take the appropriate action.

rhc54: see what you think of the latest commit - if all looks good, I'll rebase it into one commit and commit it.

Chap_API_Proc_Mgmt.tex
```diff
@@ -607,10 +607,67 @@ \section{Process Locality}
...
+The returned pointer may point to a shared memory region or an actual instance
+of the topology description. In either case, the description shall be treated
+as a "read-only" object - attempts to modify the object are likely to fail and
+return an error. The \ac{PMIx} library is responsible for performing any required cleanup when the client library finalizes.
+
+\adviceuserstart
+It is the responsibility of the user to ensure that the \refarg{topo} argument
+is properly initialized prior to calling this \ac{API}, and to check the
+returned \refarg{source} to verify that the returned topology description is
+compatible with the user's code
```
jjhursey, line 665:
```suggestion
compatible with the user's code.
```

Chap_API_Proc_Mgmt.tex
jjhursey: That sounds good

Issue #254, opened 2020-07-08
Process Set Query Clarification


jjhursey:
During the Slices/Grouping Working Group meeting on July 8, 2020 we discussed the PMIx side of the MPI Sessions prototype in Open MPI. During that conversation we noted two items that needed to be adjusted in Ch. 13 "Process Sets and Groups"
 * Remove restriction in Ch. 13 of only 1 process set per proc. Processes may be multiple process sets.
 * When querying the process set names for a given process it is possible to get either:
   - A single string if it is in exactly one process set
   - An array of strings if it is in more than one process set

jjhursey:
FYI: @rhc54 @spophale @SteVwonder @hppritcha

rhc54:
Let me ask a simple question: should we coalesce the process "set" and process "group" concepts? If we provide an API by which one can add/remove procs from a "set", then we effectively have made a "set" to be the same as a "group". The only difference between them would be these things:
```latex
* Process groups are created in collective operations, even if the collective is
   asynchronous - i.e., the group is not considered formed until _all_ members
   have actively participated in the construct operation

* Process \emph{sets} have no implied relationship between their members - i.e., a
   process in a process set has no concept of a ``pset rank'' as it would in a
   process \emph{group}

* Process \emph{groups} can be used in calls to \ac{PMIx} operations. Members of
   process \emph{groups} that are involved in an operation are translated by their
   \ac{PMIx} server into their \emph{native} identifier prior to the operation being
   passed to the host environment. For example, an application can define a process
   group to consist of ranks 0 and 1 from the host-assigned namespace of
   \emph{210456}, identified by the group id of \emph{foo}. If the application
   subsequently calls the \refapi{PMIx_Fence} \ac{API} with a process identifier of
   \{foo, PMIX_RANK_WILDCARD\}, the \ac{PMIx} server will replace that identifier with
   an array consisting of \{210456, 0\} and \{210456, 1\} - the host-assigned identifiers
   of the participating processes - prior to passing the request up to the host environment

* Process \emph{groups} can request that the host environment assign a unique
   \code{size_t} \ac{PGCID} to the group at time of group construction. An \ac{MPI}
   library may, for example, use the \ac{PGCID} as the \ac{MPI} communicator identifier
   for the group.
```
[...]

PR #258, opened 2020-07-30
Modify process set definitions

rhc54:
Per feedback, allow processes to belong to more than one process set at
a time. Clarify that process sets are defined by the host environment as
opposed to process groups that are defined by the application. Add APIs
for the host to inform the servers of changes to process sets, and event
statuses to allow the servers to communicate those changes to their
client procs.

Fixes https://github.com/pmix/pmix-standard/issues/254

### DIFF
--- a/Chap_API_Server.tex
+++ b/Chap_API_Server.tex
@@ -101,7 +101,10 @@ perform the necessary actions to scalably expose the description to the local
 clients. This includes creating any required shared memory backing stores and/
 or \ac{XML} representations, plus ensuring that all necessary key-value pairs
 for clients to access the description are included in the job-level
-information provided to each client.
+information provided to each client. All required files are to be installed
+under the effective \refattr{PMIX_SERVER_TMPDIR} directory. The \ac{PMIx}
+server library is responsible for cleaning up any artifacts (e.g., shared
+memory backing files or cached key-value pairs) at library finalize.
 \pasteAttributeItemEnd{}
 
 \optattrend
@@ -338,7 +341,7 @@ Application-level information is accessed by providing both the namespace of the
 plus the following optional information:
 
 \begin{itemize}
-    \item \pasteAttributeItem{PMIX_PSET_NAME}
+    \item \pasteAttributeItem{PMIX_PSET_NAMES}
 \end{itemize}
 
 The data may also include attributes provided by the host environment that identify the programming model (as specified by the user) being executed within the namespace. The \ac{PMIx} server library may utilize this information to customize the environment to fit that model (e.g., adding environmental variables specified by the corresponding standard for that model):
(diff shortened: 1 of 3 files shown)


rhc54:
Here is the PDF showing just the changes (highlighted): 
[pset.pdf](https://github.com/pmix/pmix-standard/files/5003057/pset.pdf)

Chap_API_Server.tex
```diff
@@ -1308,6 +1311,104 @@ \subsubsection{Initialize the cpuset structure}
...
+
+%%%%
+\descr
+
+Provide a function by which the host environment can remove processes from a
+process set. If the array of members is \code{NULL}, then all current members
+are to be removed and the process set name deleted from the \ac{PMIx} server.
+The \ac{PMIx} server shall alert all local clients of the process set
+modifications via the \refconst{PMIX_PROCESS_SET_DELETE} event.
+
```
jjhursey, line 1405: I think we should put in a line about empty sets since we have the NULL clarification. First, I guess should we allow for the existence of a set with no members? It might be a useful concept, but I don't have a use case for it. @hppritcha Does the MPI sessions proposal allow for empty psets?

If the set has 1 process and the server specifies that one process in the `members` list instead of `NULL` does the set get deleted just like if they would have passed `NULL`? Or does the server have to pass `NULL` to delete the set?

Chap_API_Sets_Groups.tex
```diff
@@ -4,38 +4,99 @@
...
     \item Process \emph{groups} can request that the host environment assign a unique \code{size_t} \ac{PGCID} to the group at time of group construction. An \ac{MPI} library may, for example, use the \ac{PGCID} as the \ac{MPI} communicator identifier for the group.
 \end{itemize}
 
-The two concepts do, however, overlap in one specific area. Process \emph{groups} are included in the process \emph{set} information returned by calls to \refapi{PMIx_Query_info_nb}. Thus, a \emph{process group} can effectively be considered an extended version of a \emph{process set} that adds dynamic definition and operational context to the \emph{process set} concept.
+The two concepts do, however, overlap in one specific area. Process
+\emph{groups} are included in the process \emph{set} information returned by
+calls to \refapi{PMIx_Query_info_nb} with the \refattr{PMIX_PSET_NAMES}.
+Thus, a \emph{process group} can effectively be considered an extended version
+of a \emph{process set} that adds dynamic definition and operational context
+to the \emph{process set} concept.
```
jjhursey, line 71: I'm not sure I like combining both progress set and progress group identifiers in the set returned by `PMIX_PSET_NAMES` because i don't know how the user would sort out which one is which. In particular if another client-side library defined a progress group then another client-side library would not be able to tell if the string returned was a process set (defined by the host) or a progress group (defined by another part of the app).

Alternatively, can we create a `PMIX_PGROUP_NAMES` that returns the process groups, and then leave `PMIX_PSET_NAMES` as just the process set names?

Chap_API_Server.tex
jjhursey: Ah - per section 13.1.1 "if all members of the set have been removed, then the set definition has been deleted from the system". I would just add that sentence here for clarity.

Chap_API_Struct.tex
```diff
@@ -4776,12 +4776,20 @@ \subsection{Query attributes}
...
-Return a comma-delimited list of the names of the psets defined in the specified range (defaults to session).
+Return a comma-delimited list of the names of the process sets defined in the
+specified range (defaults to session).
+}
+
+%
+\declareAttributeNEW{PMIX_QUERY_PSET_MEMBERSHIP}{"pmix.qry.pmems"}{pmix_proc_t*}{
+Return an array of \refstruct{pmix_proc_t} of the members of the specified
+process set.
 }
```
jjhursey, line 4793: Per my previous comment, I suggest that we create a parallel set of attributes for process groups (`PMIX_QUERY_NUM_PGROUPS`, `PMIX_QUERY_PGROUP_NAMES`, `PMIX_QUERY_PGROUP_MEMBERSHIP`) instead of combining them.

Chap_API_Struct.tex
```diff
@@ -4776,12 +4776,20 @@ \subsection{Query attributes}
...
+Return the number of process sets defined in the specified range (defaults
+to session).
 }
 
 %
 \declareAttributeNEW{PMIX_QUERY_PSET_NAMES}{"pmix.qry.psets"}{char*}{
-Return a comma-delimited list of the names of the psets defined in the specified range (defaults to session).
+Return a comma-delimited list of the names of the process sets defined in the
+specified range (defaults to session).
+}
```
jjhursey, line 4787: During the working group discussion we made this note (per https://github.com/pmix/pmix-standard/issues/254):
```
When querying the process set names for a given process it is possible to get either:
 * A single string if it is in exactly one process set
 * An array of strings if it is in more than one process set
```

I vaguely remember that we wanted an array because we wanted to avoid the limitation of reserving a special character in the string. However, returning a comma-delimited list only reservers the `,` character which might be fine. I just wanted to bring it up in case others remember better the context behind that note.

Chap_API_Server.tex
rhc54: Updated in latest commit

Chap_API_Sets_Groups.tex
rhc54: Done in latest commit

Chap_API_Struct.tex
rhc54: It gets burdensome to coders if we make the return datatype ambiguous, so let's just make it always be a pmix_data_array_t of char*, per latest commit

jjhursey: That sounds good to me.

Chap_API_Struct.tex
```diff
@@ -3586,8 +3586,8 @@ \subsection{Identification attributes}
 }
 
 %
-\declareAttributeNEW{PMIX_PSET_NAME}{"pmix.pset.nm"}{char*}{
-User-assigned name for the process set containing the given process.
+\declareAttributeNEW{PMIX_PSET_NAMES}{"pmix.pset.nm"}{char*}{
+Comma-delimited list of process sets and process groups to which the given process is a member.
```
jjhursey, line 3590: Should we split this into `PMIX_PSET_NAMES` and `PMIX_PGROUP_NAMES`? For the same reason as the query attributes.

rhc54: Yeah, and I moved them to the process set/group chapter with the rest of those related attributes.

Chap_API_Sets_Groups.tex
```diff
@@ -4,37 +4,125 @@
...
 
-Thus, process \emph{sets} differ from process \emph{groups} in several key ways:
+Host environments can create or delete process sets at any time, as well as
+add and delete members from existing process sets, through the
+\refapi{PMIx_server_define_process_set} and
+\refapi{PMIx_server_delete_process_set} \acp{API}. \ac{PMIx} servers shall
+notify all local clients of process set operations via the
+\refconst{PMIX_PROCESS_SET_DEFINE} or \refconst{PMIX_PROCESS_SET_DELETE}
+events.
```
SteVwonder, line 37: What happens to groups that are based on a given set if the set membership changes at runtime?  Are the group and set membership kept in sync automatically by PMIx?  Or is it the job of users of the groups to subscribe to set membership changes and do bookkeeping themselves? (I may be confusing the interfaces of MPI Sessions and PMIx so my apologies if this question doesn't make sense).

SteVwonder: What may help to clarify my confusion above is a better understanding of how a group is created from a set.  Is the idea that if I want to make a group based on the processes from the set `topology:/rack-1` (as a random example), that I would first query set membership with the `PMIX_QUERY_GROUP_MEMBERSHIP` key to get the process IDs in the set and then do a `PMIx_Group_construct` with that listing of process IDs?  Or is it possible to pass the set name `topology:/rack-1` directly into `PMIx_Group_construct`?

rhc54: I think the basic misunderstanding here is that groups are _not_ created from sets - groups are created from an array of proc IDs. So change in membership of one (whether a group or a set) has no impact on membership of the other, even if the user gives the group the same name as the host environment assigned to some collection of procs and both have the same procs as members.

>  I would first query set membership with the PMIX_QUERY_GROUP_MEMBERSHIP key to get the process IDs in the set and then do a PMIx_Group_construct with that listing of process IDs

Well, it would actually be `PMIX_QUERY_PSET_MEMBERSHIP`, but your method is correct. However, as stated above, that would not imply connection between the set and group.

> Or is it possible to pass the set name `topology:/rack-1` directly into `PMIx_Group_construct`?

No - remember, process sets have no operational definition, meaning that you cannot use the set name in any PMIx operation. Only process groups have operational meaning. However, you could pass that string into `PMIx_Group_construct` to name your new group `topology:/rack-1` if you wanted to "tag" the group as being based on that set - it just wouldn't mean anything to PMIx.

Does that help? Is there some change in wording required to avoid the misunderstanding? Just wondering what took you down this path.

hppritcha:
@rhc54  @jjhursey  this looks good to me.

Chap_API_Sets_Groups.tex
rhc54: Added a few words to clarify that there is no linkage between process sets and groups of the same name, even if they are formed from the same members, so changes in one aren't automatically reflected in the other.

SteVwonder: > Does that help?

Yes. That does.  Thanks for the clarifications.

> Just wondering what took you down this path.

Turns out I was getting MPI sets/groups semantics mixed up with PMIx's sets/groups. Thanks @hppritcha for reviewing the MPI sessions semantics during the functionality WG meeting.

Issue #266, opened 2020-08-13
Final prep for v4 release


rhc54:
Ralph edits - will check off each chapter once complete and ready for review after each cycle. Clear box means edits are in-progress:
- [x] Cptr 1 - Intro
- [x] Cptr 2 - Revision history
- [x] Cptr 3 - Terms
- [x] Cptr 4 - Data structs and types
- [x] Cptr 5 - Client init/fini
- [x] Cptr 6 - Sync and data access
- [x] Cptr 7 - Reserved keys
- [x] Cptr 8 - Process-related non-reserved keys
- [x] Cptr 9 - Non-process-related key-vals
- [x] Cptr 10 - Process management
- [x] Cptr 11 - Job management
- [x] Cptr 12 - Events
- [x] Cptr 13 - Data packing/unpacking
- [x] Cptr 14 - Security
- [x] Cptr 15 - Servers
- [x] Cptr 16 - Fabric
- [x] Cptr 17 - Process sets/groups
- [x] Cptr 18 - Tools
- [x] Python Binding

Josh reviews: - will check off the box when review is complete and comments are ready for re-edit. Clear box means that chapter is held for review/review is underway:
- [x] Cptr 1 - Intro
- [x] Cptr 2 - Revision history
- [x] Cptr 3 - Terms
- [x] Cptr 4 - Data structs and types
- [x] Cptr 5 - Client init/fini
- [x] Cptr 6 - Sync and data access
- [x] Cptr 7 - Reserved keys
- [x] Cptr 8 - Process-related non-reserved keys
- [x] Cptr 9 - Non-process-related key-vals
- [x] Cptr 10 - Process management
- [x] Cptr 11 - Job management
- [x] Cptr 12 - Events
- [x] Cptr 13 - Data packing/unpacking
- [x] Cptr 14 - Security
- [x] Cptr 15 - Servers
- [x] Cptr 16 - Fabric
- [x] Cptr 17 - Process sets/groups
- [x] Cptr 18 - Tools
- [x] Python Binding

rhc54:
@jjhursey I have worked thru your last set of comments and the marked chapters are ready for review. I have posted the PR mostly to make the work more visible so that others can assist in the proofreading.

rhc54:
For those wanting to assist in proofreading, I am maintaining (to the best I can) a PDF copy of the draft 
[here](https://github.com/pmix/pmix-standard/files/5071888/pmix-standard.pdf)

[...]

PR #267, opened 2020-08-13
Updates to complete PMIx v4

rhc54:
Refs #266 

Signed-off-by: Ralph H. Castain <rhc@pmix.org>


Chap_API_Sets_Groups.tex
```diff
@@ -169,63 +169,151 @@ \section{Process Groups}
...
+%
+\declareconstitemNEW{PMIX_GROUP_CONSTRUCT_ABORT}
+Any participant in a \ac{PMIx} group construct operation that returns \refconst{PMIX_GROUP_CONSTRUCT_ABORT} from the \emph{leader failed} event handler will cause all participants to receive an event notifying them of that status. Similarly, the leader may elect to abort the procedure by either returning this error code from the handler assigned to the \refconst{PMIX_GROUP_INVITE_ACCEPTED} or \refconst{PMIX_GROUP_INVITE_DECLINED} codes, or by generating an event for the abort code. Abort events will be sent to all invited or existing members of the group.
+
+%
+\declareconstitemNEW{PMIX_GROUP_CONSTRUCT_COMPLETE}
+The group construct operation has completed - the final membership will be included in the notification.
+
+%
+\declareconstitemNEW{PMIX_GROUP_LEADER_FAILED}
```
abouteiller, line 216: Why is the 'leader' feature not an implementation dependent detail?

Chap_API_Struct.tex
```diff
@@ -121,510 +125,116 @@ \subsubsection{General Error Constants}
 \declareconstitem{PMIX_ERROR}
 General Error
 %
-\declareconstitem{PMIX_ERR_SILENT}
```
abouteiller, line 124: Maybe I missed them, but a lot of  these error codes removed here do not appear to be defined somewhere else.

Chap_API_Event.tex
```diff
@@ -73,7 +73,7 @@ \subsection{\code{PMIx_Register_event_handler}}
 PMIx_Register_event_handler(pmix_status_t codes[], size_t ncodes,
                             pmix_info_t info[], size_t ninfo,
                             pmix_notification_fn_t evhdlr,
-                            pmix_evhdlr_reg_cbfunc_t cbfunc,
+                            pmix_hdlr_reg_cbfunc_t cbfunc,
```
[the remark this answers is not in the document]
rhc54: Can't help that one, I'm afraid - they will just have to update.

Chap_API_Sets_Groups.tex
```diff
@@ -169,63 +169,151 @@ \section{Process Groups}
...
-\end{constantdesc}
+\declareconstitemNEW{PMIX_GROUP_INVITE_DECLINED}
+A process has declined an invitation to join a \ac{PMIx} Group - the identifier of the declined group will be included in the notification
+
+%
+\declareconstitemNEW{PMIX_GROUP_INVITE_FAILED}
+An invited process failed or terminated prior to responding to the invitation - the identifier of the failed process will be included in the notification.
 
+%
+\declareconstitemNEW{PMIX_GROUP_MEMBERSHIP_UPDATE}
```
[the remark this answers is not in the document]
rhc54: See the MPI Sessions work.

Chap_API_Sets_Groups.tex
rhc54: This only defines the event, which has to be in the standard so it is portable - afraid I don't understand your issue.

Chap_API_Struct.tex
rhc54: They aren't because they are not returned to the user

rhc54:
Please don't comment/edit this diff - with so many people, we'll go insane trying to track all the comments and interleaving suggestions.

SteVwonder:
Thanks @rhc54 for putting this together!  I was able to skim over the first ~120 pages or so, and then skim the Tools chapter.  I left a few notes/comments in the attached PDF.  My main comment is also included below for convenience.

[pmix-standard.pdf](https://github.com/pmix/pmix-standard/files/5093823/pmix-standard.pdf)

Generally, I really like the new organization where the attributes are moved down into the relevant chapters.  That's a really nice change.  With everything being in the same chapter now, it feels like there is a lot of duplication of attribute description text.  Would it make sense to only have the full description of the attributes at the end of each chapter and in the “required/option attributes” sections just have the attribute names/links?

rhc54:
Unfortunately, all but one comment refers to chapters that I haven't "checked off" yet in #266  😄 

On the question of attribute duplicate descriptions, that's an artifact of the way @jjhursey wrote the "pasteAttributeItem" macro. I have no real issue if we decide to change that macro - I can see pros/cons about it. We can always temporarily change it as proposed and see what people think. We just shouldn't change it for v4 and then change it back later.

I'll scan and update the `PMIx_Get` attribute list.

rhc54:
I'm going to go ahead and commit this now that @jjhursey has finished his initial pass, just to set a stake in the ground for a starting point.

jjhursey:
Now that this is merged. Here is a PDF rendering of the current master for folks to review:
 * [pmix-standard.pdf](https://github.com/pmix/pmix-standard/files/5124511/pmix-standard.pdf)

PR #275, opened 2020-09-01
Clarify a cospawn pause section

jjhursey:


jjhursey:
WIP label because I want to show someone the diff before opening it for general review.

jjhursey:
The prior sentence was a bit confusing as it could be read to imply that if you colaunch then you cannot pause the application which I don't think is what we wanted.
 > Note that this method for starting the debugger precludes the use of \ac{PMIx} attributes to \emph{pause} the application prior to attaching the debugger.

I think you are correct that a better wording here might be more along the lines of:
 > Note that \ac{PMIx} job \emph{pause} attributes (e.g., \refattr{PMIX_DEBUG_STOP_IN_INIT}) do not apply to applications  where the \refattr{PMIX_DEBUGGER_DAEMONS} attribute set to true.

rhc54:
Agreed - that sounds much more accurate. Original version was poorly written by me. ☹️

jjhursey:
I pushed the updated language. I think this is ready for review now.

Chap_API_Tools.tex
```diff
@@ -749,10 +749,12 @@ \subsection{Co-Location of Debugger Daemons}
...
-target processes on the node. Note that this method for starting the debugger
-precludes the use of \ac{PMIx} attributes to \emph{pause} the application prior
-to attaching the debugger. Users are referred to the \refapi{PMIx_Job_control}
-\ac{API} for possible alternative methods.
+target processes on the node. Note that this method for starting the debugger daemons
+precludes the use of \ac{PMIx} attributes to \emph{pause} the target processes prior
+to attaching the debugger.
+Note that \ac{PMIx} job \emph{pause} attributes (e.g., \refattr{PMIX_DEBUG_STOP_IN_INIT}) do not apply to applications (defined in \refstruct{pmix_app_t}) where the \refattr{PMIX_DEBUGGER_DAEMONS} attribute is set to \code{true}.
+Users are referred to the \refapi{PMIx_Job_control}
+\ac{API} for possible alternative methods for pausing a job already in execution.
```
rhc54, line 757: Wait a minute - I got all turned around here by this blasted diff display. I thought you were addressing the "co-spawn" section. I'll address that separately as github won't let me do it here (since you didn't change it).

In this context, the application is already running - so of course you can't use the job attributes to pause it. You aren't launching the application - it is already executing, and you are simply attaching to it. Perhaps the better clarification would be:

```suggestion
target processes on the node. Note that the debugger will be attaching to the application processes
at some arbitrary point in the application's execution unless some method for pausing the application
(e.g., by providing a \ac{PMIx} directive at time of launch, or via a tool using the
\refapi{PMIx_Job_control} \ac{API} to direct that the process be paused) has been employed.
```

jjhursey: That wording does clarify it. I can move the prior wording down to the co-spawn case in this PR.

jjhursey:
PR has now been updated to clarify both the co-launch and co-spawn cases regarding the pausing of the application.

rhc54 (approved):
That got it - thanks!

PR #276, opened 2020-09-04
Adding examples

jjhursey:
* Uses the LaTeX `minted` package to include formatted source code segments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,9 +11,12 @@
 *.synctex.gz
 *.DS_Store
 *.xwm
+*.mw
 datetime-defaults.sty
 datetime.sty
 check-openpmix
 figs/~$drawings.pptx
 *.loc
 *.soc
+_minted*
+sources/_autogen_/
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 # Makefile for the PMIx Standard document in LaTex format.
 # For more information, see the master document, pmix-standard.tex.
 
+LATEX_C=pdflatex -shell-escape -file-line-error
+
 version=4.0
 default: pmix-standard.pdf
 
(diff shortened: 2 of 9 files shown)


jjhursey:
Attached is a rendering of the Examples chapter from this PR:
 * [pmix-standard-example.pdf](https://github.com/pmix/pmix-standard/files/5172218/pmix-standard-example.pdf)

There is still some cleanup to do, and I need to add some source code compile/link checking. So still a work in progress.

jjhursey:
To do list:
- [ ] If `pcc` is available then try to compile and/or link the examples.
- [ ] Investigate how we can protect the user from changing lines in the source file and it's impact on the examples in the document. Updating the line numbers.
- [ ] (Aspire) Highlight the PMIx keywords in the code examples. Requires an extension to the lexer similar to [this](https://github.com/hugomaiavieira/pygments-rspec)
- [ ] Add a Python example
- [ ] Investigate if we should use this for other C code blocks in the document.

jjhursey:
I've pushed an update to this PR that includes a preprocessor script for the examples (see latest commit). The goal here is to add comments to an example, then the preprocessor will split out the code snippets so we can include them in the latex. See `sources/hello-alt.c` for an example.

The `bin/process-example.py` script has some notes at the top on what it looks for when preprocessing the source files.

The direct minted way still works fine, but this preprocessor script should give some more flexibility.

jjhursey:
Ok this is ready for review now. See the `sources/README.md` for notes on how to use it, and [pmix-standard-example.pdf](https://github.com/pmix/pmix-standard/files/5172218/pmix-standard-example.pdf) for an example of the rendering.

jjhursey:
I just pushed a commit that adds a proper PMIx python example. Should be good now for review.

jjhursey:
For discussion today:
 * Source materials for example: https://gist.github.com/jjhursey/00b8d722567a6b71615b9e4c419edf21
 * Rendering of those sources: [pmix-standard-Examples.pdf](https://github.com/pmix/pmix-standard/files/5531879/pmix-standard-Examples.pdf)

jjhursey:
Per discussion on the Nov. 12, 2020 ASC meeting, we decided to merge this in so the working groups can use it in their work.
I rebased the branch and squashed the commits. Once CI passes then this should be ready to go.

@SteVwonder Can you give this a 👍  review so I can merge?

SteVwonder (approved):
LGTM! Thanks @jjhursey! The docker build also worked flawlessly!

PR #279, opened 2020-09-07
Update fabric chapter

rhc54:
Clarify support for multi-device scenario


jjhursey:
I was waffling about the default of `PMIX_COLLECT_GENERATED_JOB_INFO = true` when only `PMIX_COLLECT_DATA` is passed. Because it would cause implementations that are using `PMIX_COLLECT_DATA` from v3 to see a larger collective when moving to v4 without the need of the additional remote-job-local data. So to preserve behavior between the two versions it would seem that if only `PMIX_COLLECT_DATA` is passed then `PMIX_COLLECT_GENERATED_JOB_INFO` should be `false` and the end user would have to pass the extra attribute to get the extra data.

However, when considering the direct modex case, we want to `PMIx_Get` to access (and 'pull') the remote-job-local data if we are trying to access it. Further, the expectation of a `PMIx_Fence` with `PMIX_COLLECT_DATA` is that such a `PMIx_Get` operation will **not** incur an additional direct modex (that's why we did the collect in the first place).

After mulling this over a bit, I think for consistency sake between the full and direct modex operations having the default of  `PMIX_COLLECT_GENERATED_JOB_INFO = true` when only `PMIX_COLLECT_DATA` is passed to `PMIx_Fence` is the right thing to do. A client moving from v3 to v4 that does not want this additional data can add the new attribute (set to `false`) to tell the server to not include it in the data exchange. If the client didn't make this change then there is no adverse impact except that the `PMIx_Fence` with `PMIX_COLLECT_DATA` may take a little bit longer.

jjhursey (approved):

rhc54:
@jjhursey I went thru a similar thought process to reach the same conclusion.

PR #265, opened 2020-08-12
Storage Support

rhc54:
Signed-off-by: Ralph Castain <rhc@pmix.org>

PR #280, opened 2020-09-10
Storage system query support

shanedsnyder:
Based off of #265 (thanks @rhc54 )

Work in progress as we sort out `PMIX_STORAGE_TYPE` attribute.

### DIFF
--- a/Chap_API_Storage.tex
+++ b/Chap_API_Storage.tex
@@ -14,11 +14,107 @@ These attributes can be queried by applications, I/O libraries and middleware, a
 %%%%%%%%%%%
 \section{Storage support constants}
 
+\declarestruct{pmix_storage_type_t}
+
+The \refstruct{pmix_storage_type_t} structure is an enumerated type for specifying the type of storage system being used.
+
+\begin{constantdesc}
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_LUSTRE}
+Lustre storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_GPFS}
+GPFS (Spectrum Scale) storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_DATAWARP}
+Datawarp storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_NFS}
+NFS storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_CEPH}
+Ceph storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_EXT4}
+Ext4 storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_XFS}
+XFS storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_TMPFS}
+tmpfs storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_UNKNOWN_FS}
+Unknown file-based storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_DAOS}
+DAOS storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_S3}
+S3 storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_GCP}
+GCP storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_UNKNOWN_OS}
+Unknown object-based storage system.
+%
+\end{constantdesc}
+
+
+\declarestruct{pmix_storage_medium_t}
+
+The \refstruct{pmix_storage_medium_t} structure is a \code{uint16_t} type that defines a set of bit-mask flags for specifying different types of storage mediums. These can be bitwise OR'd together to accommodate storage systems that mix storage medium types.
+
+\begin{constantdesc}
+%
+\declareconstitemNEW{PMIX_STORAGE_MEDIUM_UNKNOWN}
+All bits are set to zero, indicating the storage medium type is unknown.
+%
+\declareconstitemNEW{PMIX_STORAGE_MEDIUM_TAPE}
+The storage system uses tape media.
+%
+\declareconstitemNEW{PMIX_STORAGE_MEDIUM_HDD}
+The storage system uses HDDs with traditional SAS, SATA interfaces.
+%
+\declareconstitemNEW{PMIX_STORAGE_MEDIUM_SSD}
+The storage system uses SSDs with traditional SAS, SATA interfaces.
+%
+\declareconstitemNEW{PMIX_STORAGE_MEDIUM_NVME}
+The storage system uses SSDs with NVMe interface.
+%
+\declareconstitemNEW{PMIX_STORAGE_MEDIUM_PMEM}
+The storage system uses persistent memory.
+%
+\declareconstitemNEW{PMIX_STORAGE_MEDIUM_RAM}
+The storage system is volatile (e.g., tmpfs).
+%
+\end{constantdesc}
+
+\declarestruct{pmix_storage_locality_t}
+
+The \refstruct{pmix_storage_locality_t} structure is an enumerated type for specifying the locality of storage system resources. The locality is expressed relative to the process querying the attribute.
+
+\begin{constantdesc}
+%
+\declareconstitemNEW{PMIX_STORAGE_LOCALITY_NODE}
+The storage system resources are node-local.
+%
+\declareconstitemNEW{PMIX_STORAGE_LOCALITY_NETWORK}
+The storage system resources are network-local (e.g., on the same switch).
+%
+\declareconstitemNEW{PMIX_STORAGE_LOCALITY_SITE}
+The storage system resources are site-local (i.e., in the same datacenter).
+%
+\declareconstitemNEW{PMIX_STORAGE_LOCALITY_WAN}
+The storage system resources are on the WAN (i.e., not site-local).
+%
+\end{constantdesc}
+
+
 \declarestruct{pmix_storage_access_type_t}
 
-\versionMarker{4.0}
 The \refstruct{pmix_storage_access_type_t} structure is an enumerated type for specifying storage system access types.
-All values were originally defined in version 4 of the standard unless otherwise marked.
 
 \begin{constantdesc}
 %
@@ -33,14 +129,12 @@ Provide information on storage system read and write operations.
 %
 \end{constantdesc}
 
+
 %%%%%%%%%%%
 \section{Storage support attributes}
 \label{api:struct:attributes:pstrg}
 
-\versionMarker{4.0}
-
 The following attributes may be returned in response to queries (e.g., \refapi{PMIx_Get} or \refapi{PMIx_Query_info}) made by processes or tools.
-All values were originally defined in version 4 of the standard unless otherwise marked.
 
 %
 \declareAttributeNEW{PMIX_STORAGE_ID}{"pmix.strg.id"}{char*}{
@@ -53,13 +147,23 @@ Mount point path for the storage system (valid only for file-based storage syste
 }
 
 %
-\declareAttributeNEW{PMIX_STORAGE_TYPE}{"pmix.strg.type"}{char*}{
-Qualifier indicating the type of storage being referenced by a query e.g., lustre, gpfs, online, fabric-attached, ...)
+\declareAttributeNEW{PMIX_STORAGE_TYPE}{"pmix.strg.type"}{pmix_storage_type_t}{
+Type of storage system (i.e., Lustre, GPFS, DAOS, ext4, tmpfs)
+}
+
+%
+\declareAttributeNEW{PMIX_STORAGE_MEDIUM}{"pmix.strg.medium"}{pmix_storage_medium_t}{
+Types of storage mediums utilized by the storage system (e.g., SSDs, HDDs, tape)
+}
+
+%
+\declareAttributeNEW{PMIX_STORAGE_LOCALITY}{"pmix.strg.local"}{pmix_storage_locality_t}{
+Locality of storage resources provided by the storage system (e.g., node-local or network-local)
 }
 
 %
 \declareAttributeNEW{PMIX_QUERY_STORAGE_LIST}{"pmix.strg.list"}{char*}{
-Comma-delimited list of storage identifiers (\refattr{PMIX_STORAGE_ID}) for available storage systems
+Comma-delimited list of storage identifiers (i.e., \refattr{PMIX_STORAGE_ID} types) for available storage systems
 }
 
 %
@@ -83,8 +187,13 @@ Overall used number of objects (e.g., inodes) for the storage system
 }
 
 %
-\declareAttributeNEW{PMIX_STORAGE_XFER_SIZE}{"pmix.strg.xfersz"}{uint64_t}{
-Optimal transfer size (in Kilobytes[base2]) for the storage system
+\declareAttributeNEW{PMIX_STORAGE_MINIMAL_XFER_SIZE}{"pmix.strg.minxfer"}{uint64_t}{
+Minimal transfer size (in bytes) for the storage system - this is the storage system's atomic unit of transfer (e.g., block size)
+}
+
+%
+\declareAttributeNEW{PMIX_STORAGE_OPTIMAL_XFER_SIZE}{"pmix.strg.optxfer"}{uint64_t}{
+Optimal transfer size (in bytes) for the storage system
 }
 
 %
@@ -94,7 +203,7 @@ Maximum bandwidth (in Megabytes[base2]/sec) for storage system - provided as the
 
 %
 \declareAttributeNEW{PMIX_STORAGE_BW_CUR}{"pmix.strg.bwcur"}{float}{
-Observed bandwidth (in Megabytes[base2]/sec) for storage system - provided as the most recently observed bandwidth value
+Observed bandwidth (in Megabytes[base2]/sec) for storage system - provided as a recently observed bandwidth value, with the exact measurement interval depending on the storage system and/or PMIx library implementation
 }
 
 %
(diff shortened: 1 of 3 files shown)


Chap_API_Storage.tex
```diff
@@ -0,0 +1,90 @@
...
+%%%%%%%%%%%
+\section{Storage-related attributes}
+\label{api:struct:attributes:pstrg}
+
+\versionMarker{4.0}
+
+The following attributes may be returned in response to queries (e.g., \refapi{PMIx_Get} or \refapi{PMIx_Query_info}) made by processes or tools.
+
+%
+\declareAttributeNEW{PMIX_STORAGE_ID}{"pmix.strg.id"}{char*}{
```
rhc54, line 24: I believe this is only used as a qualifier, yes? I would so indicate, if that is true.

Chap_API_Storage.tex
```diff
@@ -0,0 +1,90 @@
...
+An identifier for the storage system (e.g., lustre-fs1, daos-oss1, home-fs)
+}
+
+%
+\declareAttributeNEW{PMIX_STORAGE_PATH}{"pmix.strg.path"}{char*}{
```
rhc54, line 29: I believe this is only used as a qualifier, yes? I would so indicate, if that is true.

Chap_API_Storage.tex
```diff
@@ -0,0 +1,90 @@
...
+Mount point path for the storage system (valid only for file-based storage systems)
+}
+
+%
+\declareAttributeNEW{PMIX_STORAGE_TYPE}{"pmix.strg.type"}{char*}{
```
rhc54, line 34: This can be a qualifier, but is also something that can be returned as a value, yes? I can imagine someone asking "what is the storage type on this mount point". As written, it indicates that this can only be used as a qualifier.

Chap_API_Storage.tex
```diff
@@ -0,0 +1,90 @@
...
+
+PMIx enables users to better understand storage hierarchies by defining attributes that formalize storage system characteristics, state, and other parameters.
+These attributes can be queried by applications, I/O libraries and middleware, and workflow systems to discover available storage resources and to inform on which resources are most suitable for different I/O workload requirements.
+
+
+%%%%%%%%%%%
+\section{Storage-related attributes}
+\label{api:struct:attributes:pstrg}
+
+\versionMarker{4.0}
```
shanedsnyder, line 19: I'm not exactly sure where these `versionMarker` macros are supposed to be used, so someone might want to sanity check if it should be here, somewhere else, or just removed? Looks like they're mostly used around definitions of data structures and such, which we haven't done here, so maybe it's not needed?

Chap_API_Storage.tex
shanedsnyder: Actually, the thought I had was that you could query a PMIX_STORAGE_ID given a PMIX_STORAGE_PATH (i.e., what's the PMIX_STORAGE_ID for mount path /lustre-fs0/). But yes, it will probably be used most frequently as a qualifier.

Chap_API_Storage.tex
shanedsnyder: Same as above, though I think this one's more immediately useful -- What is the mount path corresponding to a given PMIX_STORAGE_ID? You could imagine querying PMIX_QUERY_STORAGE_LIST to get a list of storage systems PMIx is aware of, then getting their mount points so you could do I/O to them, for instance.

Chap_API_Storage.tex
shanedsnyder: Good call, I'll correct it so it's clear it's not just a qualifier, but a general attribute. 

Related to this and the 2 previous conversations, but is there a convention that should be followed to specify whether things are attributes, qualifiers, or both? Just something in the text description to help make it obvious? I was thinking it could be useful to understand what types of qualifiers are valid for different types of attributes, but don't see things organized like that elsewhere in the standard. That would help make it clear whether an attribute is something you can query, something you use to qualify, or both.

rhc54: No real convention - I just indicate it in the text if there is something unusual, e.g., that it can only be used as a qualifier. If it can be used as both a key to a query and as a qualifier, then I don't bother to say anything. I only raised the point because you had specifically called out this one attribute to be only a qualifier and wasn't sure why.

Chap_API_Storage.tex
rhc54: I wouldn't bother with it here - mostly use it for APIs.

Chap_API_Storage.tex
```diff
@@ -0,0 +1,225 @@
...
+Unknown file-based storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_DAOS}
+DAOS storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_S3}
+S3 storage system.
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_GCP}
+GCP storage system.
```
SteVwonder, line 57: I'm not familiar with GCP.  Based on some googling and proximity to the S3 type, I'm assuming this is a Google Cloud storage offering?  Would it be worth it to extend the description a bit

Chap_API_Storage.tex
```diff
@@ -0,0 +1,225 @@
...
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_UNKNOWN_OS}
+Unknown object-based storage system.
+%
+\end{constantdesc}
+
+
+\declarestruct{pmix_storage_medium_t}
+
+The \refstruct{pmix_storage_medium_t} structure is a \code{uint16_t} type that defines a set of bit-mask flags for specifying different types of storage mediums. These can be bitwise OR'd together to accommodate storage systems that mix storage medium types.
```
SteVwonder, line 67: Any reason to limit this to uint16_t as opposed to starting out with uint32_t or uint64_t?  With 7 mediums defined in this PR, it's already almost 50% allocated.

Chap_API_Storage.tex
```diff
@@ -0,0 +1,225 @@
...
+}
+
+%
+\declareAttributeNEW{PMIX_QUERY_STORAGE_LIST}{"pmix.strg.list"}{char*}{
+Comma-delimited list of storage identifiers (i.e., \refattr{PMIX_STORAGE_ID} types) for available storage systems
+}
+
+%
+\declareAttributeNEW{PMIX_STORAGE_CAPACITY_LIMIT}{"pmix.strg.caplim"}{uint64_t}{
+Overall limit on capacity (in Megabytes[base2]) for the storage system
```
SteVwonder, line 171: Storage capacity is typically given in base 10 (at least for single drives).  Would it be worthwhile to export this as just bytes to avoid the base 2 vs base 10 issue?

SteVwonder (commented):
A few inline comments below.  One general question is what error codes will be returned if the storage system doesn't know how to respond to a specific BW, Capacity, or IOPs query?

Chap_API_Storage.tex
shanedsnyder: This makes sense to me, doesn't hurt to be more descriptive in the definition.

Chap_API_Storage.tex
shanedsnyder: That's a good point, though I think we should be fine for some time. I would say the list of mediums we have defined already is exhaustive for storage devices on current systems, but always hard to speculate how much it could change going forward.

32-bits probably would have been a better decision for future-proofing, but hopefully it's not too big of a deal to up the size when new medium types are added to the standard?

Chap_API_Storage.tex
shanedsnyder: I think we decided against bytes mostly just to future-proof for exabyte storage systems (uint64_t can only represent up to 16 exabytes of storage, which is massive but we already expect OLCF Frontier to have 1 exabyte of capacity).

We chose base2 just because most of us felt that was more of the widely-used standard for storage system and operating system _software_ -- I know that's at odds with how storage hardware and network hardware/software is typically specified, but it does seem to be pretty commonplace. Hopefully being explicit in the definition is sufficient, but do agree that this inconsistency can be confusing in general.

shanedsnyder:
> A few inline comments below. One general question is what error codes will be returned if the storage system doesn't know how to respond to a specific BW, Capacity, or IOPs query?

@SteVwonder We are just assuming that storage systems and/or the PMIx server will return `PMIX_ERR_NOT_SUPPORTED` for unsupported storage system attributes similar to how different PMIx implementations can return this for unsupported function calls.

Chap_API_Storage.tex
```diff
@@ -0,0 +1,225 @@
...
+%%%%%%%%%%%
+\section{Storage support constants}
+
+\declarestruct{pmix_storage_type_t}
+
+The \refstruct{pmix_storage_type_t} structure is an enumerated type for specifying the type of storage system being used.
+
+\begin{constantdesc}
+%
+\declareconstitemNEW{PMIX_STORAGE_TYPE_LUSTRE}
```
abouteiller, line 23: I'm not a big fan of this particular enumeration. We will have to keep extending and deprecating these members as the technology background changes. Is it really critical that the value returned from a GET with `PMIX_STORAGE_TYPE` be a standardized value? (i.e., would just returning "Lustre" without specifying the corresponding constant PMIX_STORATE_TYPE_LUSTRE="Lustre" suffice?).

Some rules on capitalization, or the use of an uri form could be helpful instead?

shanedsnyder: I understand the concern with this particular enumeration. Are there existing types in the standard that are along the lines of what you describe (i.e., an attribute with values that are not explicitly defined by the standard)? Avoiding standardization of all of the different types of storage systems is certainly appealing from the standpoint of maintaining standard definitions, but on the other hand, it could affect portability of code that relies on this attribute if PMIx implementations don't adhere to the same naming conventions, etc. We could standardize some best practices for naming as you point out, but there could still be some inconsistencies in naming that could trip up users. For instance, we settled on PMIX_STORAGE_TYPE_GPFS for GPFS file systems, but GPFS is actually called now called Spectrum Scale -- you could imagine on some systems the PMIx implementation may refer to it as "GPFS", while another could refer it to as "SpectrumScale" or something like that.

shanedsnyder: If we did decide to avoid defining these types, maybe we could just use language that encouraged implementors to at least use the file system "type" contained in `mtab`. For example, on Theta this information looks like this:
```
theta-fs1 on /gpfs/theta-fs1 type gpfs (rw,relatime)
mira-home on /gpfs/mira-home type gpfs (rw,relatime)
172.22.10.92@o2ib:172.22.10.93@o2ib:/snx11214 on /lus/theta-fs0 type lustre (rw,flock,lazystatfs)
```
So, the PMIx implementation could promise to return the type (e.g., "gpfs" or "lustre") for any POSIX-like file systems, which would just leave the object storage systems (e.g., DAOS, S3) to generate names for, which is more palatable.

Chap_API_Storage.tex
SteVwonder: > hopefully it's not too big of a deal to up the size when new medium types are added to the standard?

My understanding of C's unions is pretty shaky.  What happens if the underlying value is `uint32_t`/`uint64_t` and you access it as a `uint16_t`?  Does it just truncate off the most significant digits?  If so, I think that would allow upgrading the number of bits while maintaining backwards-compatibility.

Chap_API_Storage.tex
SteVwonder, line 23: Do these types need to be coupled with version information?  I'm not sure how much it matters from an application perspective what version of NFS, GPFS, or Lustre is being used.  But I suspect it matters quite a bit for the S3 and GCP types, and for the "younger" types like DAOS.

SteVwonder: > So, the PMIx implementation could promise to return the type (e.g., "gpfs" or "lustre") for any POSIX-like file systems, which would just leave the object storage systems (e.g., DAOS, S3) to generate names for, which is more palatable.

I like that suggestion a lot.  One thought that came to mind was that OrangeFS isn't included in the current type enumeration, and I'm sure if we sat down with enough people, we'd come up with several other distributed filesystems that aren't enumerated here.  Your suggestion above handles the long-tail of filesystems quite nicely IMO.

While I'm thinking about it, would FUSE filesystems be a type that would ever be returned?

Chap_API_Storage.tex
rhc54: Changing sizes can be a problem. I would recommend just going with uint64_t from the start and avoid the issue. Given that users are accessing the values with named constants, it really doesn't make any difference, so best to just leave room.

Chap_API_Storage.tex
rhc54: > Your suggestion above handles the long-tail of filesystems quite nicely IMO.

Agreed. We'd probably want to give some thought on how to communicate to users the type "names" they should expect to get back - I'm not sure if that should be in the Standard or not? We wouldn't want it to be different across PMIx implementations.

Chap_API_Storage.tex
```diff
@@ -0,0 +1,225 @@
...
+%
+\declareconstitemNEW{PMIX_STORAGE_LOCALITY_WAN}
+The storage system resources are on the WAN (i.e., not site-local).
+%
+\end{constantdesc}
+
+
+\declarestruct{pmix_storage_access_type_t}
+
+The \refstruct{pmix_storage_access_type_t} structure is an enumerated type for specifying storage system access types.
```
rhc54, line 117: Just FYI: we avoid enumerated types in the Standard itself because it makes it impossible for local sites to extend the values. I know it isn't as nice for debuggers, but it becomes a problem when someone (e.g., a researcher) identifies some option they want to add. Implementations can implement it as an enumerated type, of course (OpenPMIx doesn't for the above reason), but we don't specify it as enumerated in the Standard.

Chap_API_Storage.tex
rhc54: We could do it as a float instead of an int...

Chap_API_Storage.tex
shanedsnyder: Sounds good, let's just change it to avoid any potential issues in the future.

Chap_API_Storage.tex
shanedsnyder: Yes, that would certainly buy us some space. Somewhat unconventional in the sense that bytes would never have a fractional value, but maybe it's better than the alternatives (not enough numerical range or having to deal with base 2 vs base 10 confusion)?

Chap_API_Storage.tex
shanedsnyder: That's a really good point. Even for things like Lustre, I'm sure it's useful for I/O middleware and other things to know what version is being used to optimize usage. So, something like this for the PMIX_STORAGE_TYPE definition (along the lines of the other thread above related to this attribute):

```
typedef struct pmix_storage_type
{
   char *type;
   char *version;
} pmix_storage_type_t;
```

Chap_API_Storage.tex
shanedsnyder: Sounds good. I used the same language for a few of the types I added, so I can just change all of them. The first is the `pmix_storage_type_t`, which I believe we are changing to a string from the above thread. The other 2 (the access_type_t you mention, and locality_t) both can just be changed to specify an integer type and then we just define the values as constants?

Chap_API_Storage.tex
shanedsnyder: Okay, so I think we want to change to a string type then (probably coupled with a version string, from @SteVwonder comment from another thread below).

As for how to communicate this information, would it make sense to include as one of the "Advice to ..." notes? I think you need to communicate to both users and implementers, though? I don't know if there's language from elsewhere in the standard we could borrow from, but we really want to encourage implementers to use this 'type' value provided by the kernel, so that users can have an expectation of what a valid type string would be.

shanedsnyder: @SteVwonder according to this: https://www.kernel.org/doc/html/latest/filesystems/fuse.html

we should be able to identify particular filesystems as being `fuse` of `fuseblk`

rhc54:
@shanedsnyder Please remember to add the "Signed-off-by: Name \<email\>" line to your commit(s) so they can be committed!

jjhursey:
ASC Meeting Oct. 1, 2020
 * Good feedback and discussion on this ticket. WG will take this back and update this PR.
 * Decided to use the "Provisional Acceptance" timeline/process for this ticket. (See governance document for exact procedure)
   - This means that the authors may continue to work on this PR and re-present it in the next quarterly meeting.
   - Well before the next meeting post another straw poll and PDF once it is ready for review.
   - Chairs will re-add the "Eligible" label and add it to the agenda.
   - In the next meeting they may present another reading and, if successful, vote.
   - If all goes well, then this would be included in the next minor release marked as "provisional"

Chap_API_Storage.tex
abouteiller: The changes that were made went the opposite direction of what I envisioned.

The removal of the actual constant values (e.g., "lustre") can cause interoperability issues between implementations and/or conversion steps between clients and servers, if the actual values varies for PMIX_STORAGE_TYPE_LUSTRE.

My initial problem was with the enumerated types for these constant values, not the fact that we do recognize some constant values as common/standardized.

Given that these are string types, we could just say that the following values are standardized: "lustre", "tmpfs", "gpfs", ... without providing the defined identifiers for them (i.e., not having #define PMIX_STORAGE_TYPE_GPFS "gpfs" in pmix.h).

jjhursey:
1Q 2021 ASC Meeting (Feb. 18, 2021)
 * Revision Exception vote passed: 10 yes / 0 no / 2 abstain / 2 absent
 * This PR enters a 2 week comment period. If no objection then it will be merged for the next minor release.

jjhursey:
It looks like some of the commits were not signed off. You will need to interactive rebase and update the commit messages. That'll like rewrite history, but that's ok. Let me know if you need a hand with that.

Once that is done then we will need to add the provisional markers (pending PR #338), but the release managers for v4 can help with that.

rhc54:
@jjhursey We can't accept this yet because there is no prototype implementation - we still have not seen the final PR for OpenPMIx. AFAIK, that is a requirement even for provisional acceptance, yes?

jjhursey:
An implementation is not required (looking at 1.3.3 of the governance doc), but anyone can request one as a reason to hold the merge. I'm ok with holding merging this until we get one.

rhc54:
Somehow, that just seems like an oversight - nobody should be able to put something in the Standard that has never been implemented, at least as a prototype. We have no way of knowing if the proposal is even feasible. Should definitely amend that section to correct this.

jjhursey:
I'm open to it. Do you want to file an issue in the governance repo for discussion?

shanedsnyder:
> It looks like some of the commits were not signed off. You will need to interactive rebase and update the commit messages. That'll like rewrite history, but that's ok. Let me know if you need a hand with that.

Just fixed this.

shanedsnyder:
FYI, what I have for initial implementation is here: https://github.com/openpmix/openpmix/pull/2086

Do we need implementation for every single attribute? Some of these aren't going to be practical in the short-term... The bandwidth and IOPS attributes obviously are going to require storage system APIs to provide the info -- Lustre has said that's something they could provide in the future, but I don't know that anything is available now. Similarly, the attributes related to storage medium and storage accessibility are going to require system-specific knowledge that I'm not sure how we'd support in OpenPMIx.

I understand the desire to want things implemented before accepting into the standard just really wish this issue was raised sooner, as we've potentially wasted a lot of time now. If we need to focus on things that can be implemented today, I suspect we're going to be really limited in what we can provide.

rhc54:
I don't think you have to implement support for everything. I just want to see that the concept has been vetted. What you posted should suffice for that purpose, and I appreciate it.

jjhursey:
I think there is a balance we need to strike with implementation requirements. Thanks Ralph for opening https://github.com/pmix/pmix-standard/issues/341 for discussion. We have a monthly call this week I'll add it to the agenda for discussion.

rhc54:
@pmix/asc-chairs  @pmix/asc-secretaries I believe this is ready to be committed, having met the review/vote as well as the implementation requirements I had requested.

Thanks @shanedsnyder for all the hard work!

abouteiller (approved):

PR #284, opened 2020-09-24
Further cleanup of the fabric and topology support

rhc54:
Rename PMIx_Fabric_update_distances to PMIx_Fabric_compute_distances and
provide parameters to support computing of distances for given
topologies. This allows support for hetero situations.

Change PMIx_Get_cpuset so it actually returns the bitmap of the current
process or thread, as specified.

Add PMIx_Parse_cpuset_string to convert a string to an actual bitmap.

### DIFF
--- a/Chap_API_Proc_Mgmt.tex
+++ b/Chap_API_Proc_Mgmt.tex
@@ -1169,6 +1169,23 @@ PMIX_TOPOLOGY_CONSTRUCT(m)
 \end{arglist}
 
 
+\littleheader{Destruct the topology structure}
+\declaremacro{PMIX_TOPOLOGY_DESTRUCT}
+
+Destruct the \refstruct{pmix_topology_t} fields
+
+\versionMarker{4.0}
+\cspecificstart
+\begin{codepar}
+PMIX_TOPOLOGY_DESTRUCT(m)
+\end{codepar}
+\cspecificend
+
+\begin{arglist}
+\argin{m}{Pointer to the structure to be destructed (pointer to \refstruct{pmix_topology_t})}
+\end{arglist}
+
+
 \littleheader{Create a topology array}
 \declaremacro{PMIX_TOPOLOGY_CREATE}
 
@@ -1184,6 +1201,24 @@ PMIX_TOPOLOGY_CREATE(m, n)
 \begin{arglist}
 \arginout{m}{Address where the pointer to the array of \refstruct{pmix_topology_t} structures shall be stored (handle)}
 \argin{n}{Number of structures to be allocated (size_t)}
+\end{arglist}
+
+
+\littleheader{Release a topology array}
+\declaremacro{PMIX_TOPOLOGY_FREE}
+
+Release a \refstruct{pmix_topology_t} array.
+
+\versionMarker{4.0}
+\cspecificstart
+\begin{codepar}
+PMIX_TOPOLOGY_FREE(m, n)
+\end{codepar}
+\cspecificend
+
+\begin{arglist}
+\arginout{m}{Address of the array of \refstruct{pmix_topology_t} structures to be released (handle)}
+\argin{n}{Number of structures in the array (size_t)}
 \end{arglist}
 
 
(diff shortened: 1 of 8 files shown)


jjhursey (dismissed):
This looks good per our conversation the other day.

rhc54:
@jjhursey Worth taking another look at - I updated the compute_distance support to handle multiple device types (e.g., GPU) and not just fabric.

Chap_API_Proc_Mgmt.tex
```diff
@@ -1285,3 +1320,345 @@ \subsection{\code{PMIx_Get_cpuset}}
...
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\subsection{Device type}
+\declarestruct{pmix_device_type_t}
+
+The \refstruct{pmix_device_type_t} is a \code{uint64_t} bitmask for identifying the type(s) whose distances are being requested, or the type of a specific device being referenced (e.g., in a \refstruct{pmix_device_distance_t} object).
+
+\versionMarker{1.0}
+\cspecificstart
+\begin{codepar}
+typedef uint16_t pmix_device_type_t;
```
jjhursey, line 1514: I like where this is going, but have a suggestion.

What if a system has multiple types of GPUs, block devices, and networks - and we just want the distances to the specific subset? You kind of already hit this with network distinguishing between Ethernet vs OpenFabrics. I could also see wanting only the distances to the NVMe devices on the node and not the other storage devices for local checkpoint storage.

In the `endpoint_t` we return the `osname` of the device, maybe we can make this a slightly more complex type to incorporate something similar. Meaning:
```
typedef uint16_t pmix_device_type_class_t;

typedef struct pmix_device_type \{
    char *osname;
    pmix_device_type_class_t type;
\} pmix_device_type_t;
```

The `type_class_t` would be a single value for the class of the device. If `osname` is `NULL` then it works with all devices of that class, if `osname` is not NULL then that value is used to select the subset of devices that match that `osname` in this class. Maybe `osname` isn't the correct term, but you get the idea of what I'm talking about.
Then in the other APIs that consume this type they would take an array of this type.

For the network example, you would specify `{PMIX_DEVTYPE_NETWORK, "openfabrics"}` to get just the openfabrics devices. The value passed here is implementation-defined, but the keys (device classes) are standard-defined.

Actually, instead of adding a new data structure we could just add an attribute to the APIs with standard-defined keys for the device class, and values defined by the implementation. Then the APIs just need an info array argument to specify the device target(s).

rhc54: I confess I had to re-read this several times to really grok it, but I think I caught the gist. I have updated the `PMIx_Compute_distances` APIs to take an array of info instead of the `pmix_device_type_t` argument so we have more flexibility. See what you think.

jjhursey: I think that's fine. I guess I was thinking that we could eliminate `pmix_device_type_t `, convert these constants to attributes, and then use some more generic attribute/value pairs to specify the device type.

So we would have the attributes:
 * `PMIX_DEVTYPE_BLOCK` with values like "sda,sdb" or "nvme"
 * `PMIX_DEVTYPE_COPROC` with values that are implementation-specific for "gpu", "mic". The user could be even more specific if they just want a subset of the devices "mic0,mic1"
 * `PMIX_DEVTYPE_NETWORK` with values like "ethernet", "openfabrics". Or more specific "eth0,eth1" or "mlx5_0,mlx5_1"

Then the set of attributes can be a bit more general, but the values can as specific as the user needs them to be. They can specify multiple attributes in the info keys if they want multiple device types. Instead of a comma-separated value like "eth0,eth1" we could also say that the value is a single item, and if they want to specify more than one value they duplicate the attribute in the info key array.

rhc54: I was trying to avoid the "magic word" problem - e.g., making users "know" that the right word is "ethernet" for the attribute value to get what they want. If we move to what you propose, it seemed to me that we also then need a list of string constants (e.g., `PMIX_OPENFABRICS` = "pmix.openfab") to help people specify something the implementation can understand. Once you take that step, you're pretty much right back to this PR, with string constants instead of numerical ones.

jjhursey: That's a fair point. (Similar to what the storage folks were talking about today) I'm ok with leaving this like it is here.

PR #286, opened 2020-10-01
Remove "draft" designation

rhc54:
Builds on https://github.com/pmix/pmix-standard/pull/285


jjhursey (changes requested):
I don't think we want to remove the draft designation on the branch. By leaving the draft designation anyone that clones the `v4` branch to create a change request will automatically generate draft versions of the standard. If we removed the draft designation then the proposer would have to intentionally change it back to draft for before generating a PDF to share and submitting the PR for review.

rhc54:
Good point - I'll leave it to you to determine how we generate the final document.

PR #292, opened 2020-10-03
Updates to Python bindings and missed API

rhc54:
Replace the bitarray types in Python bindings with simple integer lists
for ease of use/translation. Cleanup some missing functions.

Add PMIX_EXTERNAL_PROGRESS attribute and PMIx_Progress API so that a
host can externally progress the PMIx library. This is less onerous than
requiring the host to also adopt libevent and provide an event base.
Deprecate the PMIX_EVENT_BASE attribute.

I had forgotten about the external progress promise until someone
pointed it out to me.


rhc54:
Let's bring these in so folks can see the current status here. I believe there will probably be one more iteration before release.

PR #294, opened 2020-10-13
Final touchups to PMIx_Get and PMIx_server_init attributes

rhc54:
Provide the ability to request pointers to values, filling in of a
provided value struct address, as well as the default return of a newly
malloc'd struct. Clarify that "success" means the data was returned in
the form requested - do not return it as something different.

Remove the "slot" definition per WG discussion


rhc54:
@jjhursey I believe this completes v4!!

jjhursey (approved):
This looks good. I like the new attribute name for returning a pointer to the data. That makes it more clear what's happening.

PR #303, opened 2020-10-28
Minor cleanups and clarifications

rhc54:
Add reserved prefixes for regular expressions and explain that such
prefixes are required to be present. Expand on explanation of refresh
cache behavior and required keys. Add new compressed byte object data
type.


Chap_API_Server.tex
```diff
@@ -652,13 +652,21 @@ \subsubsection{Assembling the registration information}
...
-Changing the filter criteria allows the construction of node maps for any level of information.
+Changing the filter criteria allows the construction of node maps for any level of information. The returned regular expression shall start with a string identifier of the method used to generate the expression. This is followed by a colon (':') delimiter and then the regular expression itself. The following identifiers are reserved by the Standard:
+
+\begin{itemize}
+    \item \emph{raw} - indicates that the expression is simply the comma-delimited input string (no processing was performed)
+    \item \emph{native} - a \ac{PMIx}-unique regular expression generator and parser
+    \item \emph{compress} - a compressed format based on the zlib compression package
+\end{itemize}
+
+Note that the regular expression itself is not required to be a printable string nor to obey typical string constraints (e.g., include a \code{NULL} terminator byte).
```
jjhursey, line 669: We need to update the "Advice to PMIx server hosts" in both the `PMIx_generate_regex` and `PMIx_generate_ppn` functions (16.2.1 and 16.2.2). That advice is where we introduce the colon delimiter with either the `pmix:\0` prefix or the `blob:\0` prefix. Note that we added the `:\0` (with the NUL character) so that if the user tired to print it they wouldn't accidentally be printing an unterminated string. I think that it would be good to preserve that.

From this addition it seems like `native: == pmix:` and `compress: == blob:`. Why do we need the name change?

I think this text should be moved to `PMIx_generate_regex` (with a reference to it in `PMIx_generate_ppn`) as normal text since it is putting a requirement on those two functions. Then trim up the Advice blocks since some of that is normative text now.

In this particular section, it would be good to keep a sentence or two about the regex nature of the returned "string" specifically indicating that the return value is not necessarily printable (maybe keep your last sentence here).

jjhursey (changes requested):
A note on the placement of the new regex prefixes WRT the old prefixes that needs to be resolved. Otherwise, the rest of the PR looks fine to me.

Chap_API_Server.tex
rhc54: Sigh - just shows you shouldn't try to write documents when half asleep and multi-tasking on a telecon 😄 I just _knew_ I had written the prefix stuff up somewhere, but couldn't find it!

I'll do as suggested and push the change a little later - after I have completely woken up this time!

rhc54: @jjhursey Okay, see what you think now!

Chap_API_Server.tex
```diff
@@ -273,13 +273,15 @@ \subsection{\code{PMIx_generate_regex}}
 
 Given a comma-separated list of \refarg{input} values, generate a reduced size representation of the input that can be passed down to the \ac{PMIx} server library's \refapi{PMIx_server_register_nspace} \ac{API} for parsing. The order of the individual values in the \refarg{input} string is preserved across the operation. The caller is responsible for releasing the returned data.
 
-The precise compressed representations will be implementation specific. However, all \ac{PMIx} implementations are required to include a \code{NULL}-terminated string in the output representation that can be printed for diagnostic purposes.
+\label{regex:fmt}The precise compressed representations will be implementation specific. The regular expression itself is not required to be a printable string nor to obey typical string constraints (e.g., include a \code{NULL} terminator byte). However, all \ac{PMIx} implementations are required to include a colon-delimited \code{NULL}-terminated string at the beginning of the output representation that can be printed for diagnostic purposes and identifies the method used to generate the representation. The following identifiers are reserved by the Standard:
```
jjhursey, line 276:
```suggestion
\label{regex:fmt}The precise compressed representations will be implementation specific. The regular expression itself is not required to be a printable string nor to obey typical string constraints (e.g., include a \code{NULL} terminator byte). However, all \ac{PMIx} implementations are required to include a colon-delimited \code{NULL}-terminated string at the beginning of the output representation that can be printed for diagnostic purposes and identifies the method used to generate the representation. The following identifiers are reserved by the \ac{PMIx} Standard:
```

Chap_API_Server.tex
```diff
@@ -273,13 +273,15 @@ \subsection{\code{PMIx_generate_regex}}
...
 
-\advicermstart
-The returned representation may be an arbitrary array of bytes as opposed to a valid \code{NULL}-terminated string. However, the method used to generate the representation shall be identified with a colon-delimited string at the beginning of the output. For example, an output starting with \code{"pmix:\textbackslash0"} might indicate that the representation is a \ac{PMIx}-defined regular expression represented as a \code{NULL}-terminated string following the \code{"pmix:\textbackslash0"} prefix. In contrast, an output starting with \code{"blob:\textbackslash0"} might indicate a compressed binary array follows the prefix.
+\begin{itemize}
+    \item \code{"raw:\textbackslash0"} - indicates that the expression following the identifier is simply the comma-delimited input string (no processing was performed).
+    \item \code{"pmix:\textbackslash0"} - a \ac{PMIx}-unique regular expression represented as a \code{NULL}-terminated string following the identifier.
+    \item \code{"blob:\textbackslash0"} - a compressed binary array format based on the \emph{zlib} compression package, with the size encoded in the space immediately following the identifier.
```
jjhursey, line 281:
```suggestion
    \item \code{"blob:\textbackslash0"} - a \ac{PMIx}-unique regular expression that is not represented as a \code{NULL}-terminated string following the identifier. Additional implementation-specific metadata may follow the identifier along with the data itself. For example, a compressed binary array format based on the \emph{zlib} compression package, with the size encoded in the space immediately following the identifier.
```

What do you think about this? Then the `blob` can be the zlib compression, or anything the implementation wants to pack in there (e.g., using a different compression library/algorithm than zlib). The PMIx implementation could encode it as `blob:\0compoent:zlib\0size:1234\0...` for zlib. But the `blob` indicates that it is not a traditionally printed string after the identifier. It could be a string, but with additional metadata that the implementation knows how to parse or it could be a binary format of some sort.

jjhursey (commented):
This is looking good. I had 1 minor item, and 1 suggested change for discussion.

jjhursey (approved):
👍

PR #305, opened 2020-11-04
Some clarifications on tool indirect launch

rhc54:
Working thru this section and cleaning up based on implementation

### DIFF
--- a/Chap_API_Tools.tex
+++ b/Chap_API_Tools.tex
@@ -91,21 +91,23 @@ The rendezvous methods are automatically employed for the initial tool connectio
 
 Connection to a local \ac{PMIx} server is pursued according to the following precedence chain based on attributes contained in the call to the \refapi{PMIx_tool_init} or \refapi{PMIx_tool_connect_to_server} \acp{API}. Servers to which the tool already holds a connection will be ignored. Except where noted, the \ac{PMIx} library will return an error if the specified file cannot be found, the caller lacks permissions to read it, or the server specified within the file does not respond to or accept the connection — the library will not proceed to check for other connection options as the user specified a particular one to use.
 
+Note that the \ac{PMIx} implementation may choose to introduce a "delayed connection" protocol between steps in the precedence chain - i.e., the library may cycle several times, checking for creation of the rendezvous file each time after a delay of some period of time, thereby allowing the tool to wait for the server to create the rendezvous file before either returning an error or continuing to the next step in the chain.
+
 \begin{itemize}
 %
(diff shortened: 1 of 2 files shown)


Chap_API_Tools.tex
```diff
@@ -1338,6 +1344,57 @@ \subsection{\code{PMIx_tool_set_server}}
...
 
 
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\subsection{\code{PMIx_tool_check_connection}}
+\declareapi{PMIx_tool_check_connection}
+
+%%%%
+\summary
+
+Check if a specified process has connected with the requesting tool.
```
jjhursey, line 1354: I'm not sure I follow this description. The API indicates that the calling tool is checking its connection with the specified server.

I assume that they still need to call `PMIx_tool_attach_to_server` to connect with this server. Or `PMIx_tool_set_server` to switch the connection to this server. The former, at least, should not return until there is a successful connection. What is the use case for this new API then?

rhc54: Imagine the case where a tool fork/exec's an intermediate launcher (IL). In this case, the tool needs to know when the IL has connected back to the tool so the tool can set the IL as its server and then issue directives such as calling `PMIx_Spawn`. Ideally, it would be nice to encapsulate that inside the tool's call to `PMIx_Spawn` as the Standard currently reads. However, that is proving difficult without creating another progress thread. This is an alternate solution.

I suppose another option is to move the delay for connection into "set_server". Let me try and see if there is a way to make that work. We still might want this API as a way of checking the state of a connection (since a tool doesn't have the server's upcall module functions to tell it when a server disappears), but we can hold on that for now.

jjhursey: Adding the delay to `PMIx_tool_attach_to_server` makes sense to me - to wait for the designated server to come online. Additionally adding it to `PMIx_tool_set_server` to wait for the confirmation from the server seems good too since we might be switching to this server, and we need to make sure it is still alive.

The new API might be useful if you want to do a liveness check on the server, but I would expect it to return a boolean state given its name (`PMIx_tool_check_connection`). If the server disappears we should get an event in the tool (if we registered for it), right? I would expect the Tool-side PMIx library to throw that event.

rhc54:
@jjhursey I believe this is ready now. I removed the new API as it is not necessary, as we discussed.

jjhursey (approved):
I think this looks good.

jjhursey:
@rhc54 Is this ready to merge? I assume so but thought I would double-check.

rhc54:
Yeah - just been busy. Will merge now.

PR #306, opened 2020-11-05
Update release dates for 3.2 and 4.0

jjhursey:
Signed-off-by: Joshua Hursey <jhursey@us.ibm.com>


rhc54 (commented):
I would suggest just making v4 be in Dec - probably more realistic at this point. I have to complete the tools work and then still need to do the Python bindings.

jjhursey:
Per PMIx teleconf discussion. Let's keep v4 as Nov for now, and if it slips to Dec then we can update it then.

Issue #307, opened 2020-11-19
When exactly to return `PMIX_ERR_NOT_SUPPORTED` in `pmix_server_abort_fn_t`


krempel-pt:
Hello,

it is not clear to me for every case, how the host server using the support library needs to behave when `pmix_server_module_t.abort` is called.

If everything is supported, it is quiet clear:
- If `procs == NULL`, terminate all processes of the calling client's namespace and return `PMIX_OPERATION_SUCCEEDED` or directly return `PMIX_SUCCESS` and after terminating them call the callback function.
- If `procs != NULL` terminate the processes in the array and return `PMIX_OPERATION_SUCCEEDED` or directly return `PMIX_SUCCESS` and after terminating them call the callback function.

But there are some open questions:
- If `procs != NULL` but the host server is not able or not configured to terminate a subset of a namespace. Should it then return `PMIX_ERR_NOT_SUPPORTED` and do nothing or is it standard conform, too, to behave like `procs == NULL`, terminate the whole namespace and return PMIX_SUCCESS? 

This could be clarified in 16.3.5 as "Advice to PMIx server hosts" paragraph on when to return `PMIX_ERR_NOT_SUPPORTED` and which decisions on behavior are up to the host server.

- Is it allowed, that `procs` contains processes that are not in the namespace of the process calling `PMIx_Abort`?

This could be clarified in 11.1.1 or in 16.3.5 or both.

### References
4.0 draft:
- 16.3.5 `pmix_server_abort_fn_t`
- 11.1.1 `PMIx_Abort`

rhc54:
You raise some good points. I wrestled with two possible responses for the edge cases you presented:

* return `PMIX_ERR_PARTIAL_SUCCESS` and terminate the procs that the host can terminate. I didn't really feel comfortable with this option as there is no way for the host to inform the caller as to which procs it terminated, and which are still executing.

* return `PMIX_ERR_BAD_PARAM` to indicate that the API itself is supported, but that the caller provided a parameter that lies outside of the ability of the host to support. In this case, the host would take no action - i.e., no processes would be aborted.

The second option appeared the better one to me, and so that is what I have proposed in the #308. I'd welcome your comments on that RFC.

jjhursey:
It seems that the server should return `PMIX_ERR_NOT_SUPPORTED` and take no action in this case. The parameter is fine t pass to the API, but just not supported by the RM.

rhc54:
[...]

PR #308, opened 2020-11-20
Clarify the PMIx_Abort behavior

rhc54:
Deal with the edge cases of environments where not all use-cases are
supported (e.g., termination of a subset of a namespace, or termination
of processes outside of the caller's namespace)

Refs #307


Chap_API_Struct.tex
```diff
@@ -169,7 +169,7 @@ \subsection{PMIx Return Status Constants}
 suitable connection either has not been or can not be made.
 %
 \declareconstitem{PMIX_ERR_BAD_PARAM}
-An incorrect parameter was passed to a \ac{PMIx} \ac{API}.
+An incorrect or unsupported parameter was passed to a \ac{PMIx} \ac{API}.
```
jjhursey, line 172: This gets confusing when compared with the `PMIX_ERR_NOT_SUPPORTED` error code. If an unsupported parameter was passed then it should

jjhursey: Since we have `PMIX_ERR_PARAM_VALUE_NOT_SUPPORTED` now, I don't think we want this change to the `PMIX_ERR_BAD_PARAM` definition.

jjhursey (changes requested):
Looks good other than my request to revert the definition of `PMIX_ERR_BAD_PARAM`.

rhc54:
> Looks good other than my request to revert the definition of `PMIX_ERR_BAD_PARAM`.

Yeah, I'm noodling over that one. Obviously, a "bad param" returned error doesn't mean you passed an incorrect type to us - the compiler would have to catch that one. We have now separated out the issue of passing a supported parameter with an unsupported value, so that isn't included here.

Seems to me that the other cases we need to cover are:

* passing a given attribute more than once, with non-matching values
* passing multiple attributes that result in conflicting directives
* passing an attribute that this function does not support/recognize, at least in this implementation and/or environment

I think we return `PMIX_ERR_NOT_SUPPORTED` for the last one, which should be fine. I think we should use `PMIX_ERR_BAD_PARAM` for the first two.

Does that make sense? If so, it might be worth adding some verbage to convey those cases.

jjhursey:
> > Looks good other than my request to revert the definition of `PMIX_ERR_BAD_PARAM`.
> 
> Yeah, I'm noodling over that one. Obviously, a "bad param" returned error doesn't mean you passed an incorrect type to us - the compiler would have to catch that one. We have now separated out the issue of passing a supported parameter with an unsupported value, so that isn't included here.

I think so. Could we say `An incorrect, conflicting, or invalid parameter was passed...`?

> * passing a given attribute more than once, with non-matching values

This would be a "conflicting" case for `PMIX_ERR_BAD_PARAM`

> * passing multiple attributes that result in conflicting directives

This would be a "conflicting" case for `PMIX_ERR_BAD_PARAM`

> * passing an attribute that this function does not support/recognize, at least in this implementation and/or environment

I would think this would be a `PMIX_ERR_NOT_SUPPORTED`

rhc54:
Sounds like we are on the same wavelength - I'll try to provide an update in a bit.

rhc54:
@jjhursey See what you think of the revision to `PMIX_ERR_BAD_PARAM`

jjhursey (approved):
This looks good to me

PR #315, opened 2020-12-13
Update tools chapter

rhc54:
Give more details on the indirect launch case, including separation of
the case where the tool knows how to parse its cmd line to extract the
launcher vs any applications, and the case where the tool does NOT know
how to parse the cmd line.

### DIFF
--- a/Chap_API_Tools.tex
+++ b/Chap_API_Tools.tex
@@ -72,7 +72,7 @@ Tool process identifiers are assigned by one of the following methods:
         \item If \refattr{PMIX_TOOL_RANK} is also given, then the rank of the tool will be assigned that value.
         \item If \refattr{PMIX_TOOL_RANK} is not given, then the rank will be set to a default value of zero.
     \end{itemize}
-    \item If a process ID has not been provided and the tool connects to a server, then one will be assigned by the host environment upon connection to that server.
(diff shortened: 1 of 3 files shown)


Chap_API_Tools.tex
```diff
@@ -72,7 +72,8 @@ \section{Connection Mechanisms}
         \item If \refattr{PMIX_TOOL_RANK} is also given, then the rank of the tool will be assigned that value.
         \item If \refattr{PMIX_TOOL_RANK} is not given, then the rank will be set to a default value of zero.
     \end{itemize}
-    \item If a process ID has not been provided, then one will be assigned by the host environment upon connection to a server. Users should note that the tool's process ID will be \emph{invalid} until a connection has been established.
+    \item If a process ID has not been provided and the tool connects to a server, then one will be assigned by the host environment upon connection to that server.
```
jjhursey, line 75:
```suggestion
    \item If a process ID is not provided and the tool connects to a server, then one will be assigned by the host environment upon connection to that server.
```
Just to match the wording of the point that follows.

jjhursey (approved):
I have one minor wording consistency fix, but other than that this is a really good addition to the explanation of the indirect cases.

PR #322, opened 2020-12-22
Add a directive indicating that a reqd attribute has been processed

rhc54:
Each level (client, server, host) needs to know if some prior level has
already processed a required attribute so it can correctly deal with it.
Define an info directive for that purpose, and provide macros to set and
test for that directive.


Chap_API_Struct.tex
```diff
@@ -1455,6 +1455,9 @@ \subsection{Info Type Directives}
 \declareconstitem{PMIX_INFO_REQD}
 The behavior defined in the \refstruct{pmix_info_t} array is required, and not optional. This is a bit-mask value.
 %
+\declareconstitemNEW{PMIX_INFO_REQD_PROCESSED}
+Mark that this required attribute has been processed. A required attribute can be handled at any level - the\ac{PMIx} client library might take care of it, or it may be resolved by the \ac{PMIx} server library, or it may pass up to the host environment for handling. If a level does not recognize or support the required attribute, it is required to pass it upwards to give the next level an opportunity to process it. Thus, the host environment (or the server library if the host does not support the given operation) must know if a lower level has handled the requirement so it can return a \refconst{PMIX_ERR_NOT_SUPPORTED} error status if the host itself cannot meet the request. Upon processing the request, the level must therefore mark the attribute with this directive to alert any subsequent levels that the requirement has been met.
```
jjhursey, line 1459: Minor whitespace issue.
```suggestion
Mark that this required attribute has been processed. A required attribute can be handled at any level - the \ac{PMIx} client library might take care of it, or it may be resolved by the \ac{PMIx} server library, or it may pass up to the host environment for handling. If a level does not recognize or support the required attribute, it is required to pass it upwards to give the next level an opportunity to process it. Thus, the host environment (or the server library if the host does not support the given operation) must know if a lower level has handled the requirement so it can return a \refconst{PMIX_ERR_NOT_SUPPORTED} error status if the host itself cannot meet the request. Upon processing the request, the level must therefore mark the attribute with this directive to alert any subsequent levels that the requirement has been met.
```

jjhursey (approved):
This looks good. I found one minor whitespace issue you can fix and sqash. Then it's good to go.

PR #325, opened 2021-01-06
Remove stale language and roll to v4.1

### LINKED ISSUE #324: Where does evhdlr_ref argument to PMIx_Degregister_event_handler come from?
### Overview

Event handlers are registered with the following call:

pmix_status_t PMIx_Register_event_handler(pmix_status_t codes[], size_t ncodes,
  pmix_info_t info[], size_t ninfo, pmix_notification_fn_t evhdlr, pmix_hdlr_reg_cbfunc_t cbfunc, void *cbdata);

This returns PMIX_SUCCESS on successful completion.  The cbfunc/cbdata can be NULL if the function is treated as blocking.  Doesn't seem to return any handle to the event.

However, deregistering events is done with:

pmix_status_t PMIx_Deregister_event_handler(size_t evhdlr_ref, pmix_op_cbfunc_t);

which takes a size_t evhdlr_ref.  I have no idea where to get this value.

I looked for examples and the only ones I could find (https://github.com/openpmix/openpmix/blob/master/examples/server.c) just used a hard-coded 0 for the evhdlr_ref, which seems if it works, surely only works in the case where you have a single event handler registered.

The text on the 2nd page of the event notification chapter also says:

"Ordering can be specified either by providing the relevant returned event handler registration ID or  using event handler names, if the user specified an event handler name when registering the corresponding event."

This really makes me think I'm missing something obvious here, but then again... the event handler registration ID is a size_t and the user-specified name is a (char*), so you couldn't possibly specify the registration id to something like PMIX_EVENT_HDLR_BEFORE.  This makes me think it doesn't really exist.

### Motivation

Maybe broken API?

### Discussion Items

Am I missing something?
If this is broken, how to fix?  
I don't see how the existing registration API can be used to return this reference... even if we wanted to change the meaning of the return value (which would be really bad), the types wouldn't work out.  The info's can't be used to return values.   It really seems like the API needs an additional parameter?

rhc54:
Remove stale language
The event handler registration ordering is based on event handler names, not event handler reference IDs

Closes #324 

-----------------------

Roll the version to v4.1

-----------------------


jjhursey:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

jjhursey:
This is an errata item, and we can use the provisional process for it (though I don't think we need to mark it as provisional in the doc as that gets messy). This will be presented at the 1Q 2021 ASC meeting, and (unless there are objections) will be committed at that time.

jjhursey (approved):

jjhursey:
The Co-Chairs miss this PR when compiling the 1Q 2021 agenda (Sorry). It will be presented in the plenary during 1Q2021, and is up for inclusion in 2Q2021.

pmix-standard.tex
```diff
@@ -57,7 +57,7 @@
 
 % Text to appear in the footer on even-numbered pages:
 \ifthenelse{\boolean{is_unofficial_draft}}
-  {\newcommand{\VER}{4.0 (Draft)}
+  {\newcommand{\VER}{4.1 (Draft)}
```
abouteiller, line 60:
```suggestion
  {\newcommand{\VER}{4.0 (Draft)}
```

abouteiller (commented):
During the 21Q2 ASC meeting, it was identified that this change should happen in another PR.

jjhursey:
 * Passed first vote: 7 yes / 0 no / 1 abstain
 * This has been accepted as provisional.
 * I will do the following today:
   * Update the branch
   * Move the version change to a separate PR per [this comment](https://github.com/pmix/pmix-standard/pull/325#pullrequestreview-659065420)
   * Merge it into `v4`
   * Cherry-pick to `master`

jjhursey (approved):

jjhursey:
* Update the branch to `v4` HEAD
   * Revert the version change to a separate PR per [this comment](https://github.com/pmix/pmix-standard/pull/325#pullrequestreview-659065420)
   * Merge it into `v4`
   * No other text changes made

PR #328, opened 2021-01-14
Add appendix with use case descriptions

SteVwonder:
Introduces the `\copySignature` and `\pasteSignature` macros for easy copying and pasting of function signatures across the document, along with a `convert-to-copysig.py` script that can convert old-style signatures to the new `\copySignature` format (idempotently).

The new appendix contains use case documents for:

- Business Card Exchange (#191)
- Debugging/Tools (#216)
- Hybrid Programming Models (#232)

Co-authored by @spophale and @jjhursey

[PDF copy with the changes](https://github.com/pmix/pmix-standard/files/5839436/pmix-standard.pdf) (updated Jan 19, 2021)

PR #330, opened 2021-01-29
Missing v4 patches

jjhursey:
* master version of PR #315 that went into v4
 * With this master and v4 are in sync.


jjhursey:
I found this mistake while working on the debugger use case for PR #328. I noticed master was missing an attribute (`PMIX_LAUNCH_DIRECTIVES`) that didn't seem to be on master, but was on v4. I tracked it down to PR #315 that we must have forgotten to cherry-pick back to master. Now a diff of master and v4 only show the changes merged to master since v4 was finalized.

rhc54 (approved):
Good catch!

PR #332, opened 2021-01-30
Add four missing data management functions

rhc54:
While we provided macros for these purposes, it turns out that the logic in implementing those macros is somewhat more complex that originally realized and better suited to an actual function call. We also missed functions for compressing/decompressing data. This RFC adds to functions in the Data Management chapter:

- PMIx_Data_load transfers a data "blob" from a pmix_byte_object_t structure to a pmix_data_buffer_t

- PMIx_Data_unload transfers the payload in a pmix_data_buffer_t to a pmix_byte_object_t structure
- PMIx_Data_compress applies a lossless data compression algo to the input data
- PMIx_Data_decompress decompresses the input data (which must have been compressed by PMIx_Data_compress)

### DIFF
--- a/Chap_API_Data_Mgmt.tex
+++ b/Chap_API_Data_Mgmt.tex
@@ -467,7 +467,7 @@ The buffer must be allocated or constructed in advance - failing to do so
 will cause the load function to return an error code or fail.
 
 The caller is responsible for pre-packing the provided
-payload - the load function cannot convert to network byte order
+payload. For example, the load function cannot convert to network byte order
 any data contained in the provided payload.
 \adviceuserend
 
--- a/Chap_API_Data_Mgmt.tex
+++ b/Chap_API_Data_Mgmt.tex
@@ -539,7 +539,7 @@ Perform a lossless compression on the provided data
 \cspecificstart
 \begin{codepar}
 bool
-PMIx_Data_compress(uint8_t *inbytes, size_t size,
+PMIx_Data_compress(const uint8_t *inbytes, size_t size,
                    uint8_t **outbytes, size_t *nbytes);
 \end{codepar}
 \cspecificend
--- a/Chap_API_Data_Mgmt.tex
+++ b/Chap_API_Data_Mgmt.tex
@@ -421,3 +421,105 @@ This function will append a copy of the payload in one buffer into another buffe
 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\subsection{\code{PMIx_Data_load}}
+\declareapi{PMIx_Data_load}
+
+%%%%
+\summary
+
+Load a buffer with the provided payload
+
+%%%%
+\format
+
+\versionMarker{4.1}
+\cspecificstart
+\begin{codepar}
+pmix_status_t
+PMIx_Data_load(pmix_data_buffer_t *dest,
+               pmix_byte_object_t *src);
+\end{codepar}
+\cspecificend
+
+\begin{arglist}
+\argin{dest}{Pointer to the destination \refstruct{pmix_data_buffer_t} (handle)}
+\argin{src}{Pointer to the source \refstruct{pmix_byte_object_t} (handle)}
+\end{arglist}
+
+Returns one of the following:
+\begin{constantdesc}
+\item \refconst{PMIX_SUCCESS} The data has been loaded as requested
+\item \refconst{PMIX_ERR_BAD_PARAM} The \refarg{dest} structure pointer is \code{NULL}
+\item \refconst{PMIX_ERR_NOT_SUPPORTED} The \ac{PMIx} implementation does not support this function.
+\end{constantdesc}
+
+%%%%
+\descr
+
+The load function allows the caller to transfer the contents of the \refarg{src}
+\refstruct{pmix_byte_object_t} to the \refarg{dest} target buffer. If a payload
+already exists in the buffer, the function will "free" the existing data to
+release it, and then replace the data payload with the one provided
+by the caller.
+
+\adviceuserstart
+The buffer must be allocated or constructed in advance - failing to do so
+will cause the load function to return an error code or fail.
+
+The caller is responsible for pre-packing the provided
+payload - the load function cannot convert to network byte order
+any data contained in the provided payload.
+\adviceuserend
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\subsection{\code{PMIx_Data_unload}}
+\declareapi{PMIx_Data_unload}
+
+%%%%
+\summary
+
+Unload a buffer into a byte object
+
+%%%%
+\format
+
+\versionMarker{4.1}
+\cspecificstart
+\begin{codepar}
+pmix_status_t
+PMIx_Data_unload(pmix_data_buffer_t *src,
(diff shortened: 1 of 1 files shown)


rhc54:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

Chap_API_Data_Mgmt.tex
```diff
@@ -420,4 +420,106 @@ \subsection{\code{PMIx_Data_copy_payload}}
...
+
+The load function allows the caller to transfer the contents of the \refarg{src}
+\refstruct{pmix_byte_object_t} to the \refarg{dest} target buffer. If a payload
+already exists in the buffer, the function will "free" the existing data to
+release it, and then replace the data payload with the one provided
+by the caller.
+
+\adviceuserstart
+The buffer must be allocated or constructed in advance - failing to do so
+will cause the load function to return an error code or fail.
```
jjhursey, line 467:
```suggestion
will cause the load function to return an error code.
```

Chap_API_Data_Mgmt.tex
```diff
@@ -420,4 +420,106 @@ \subsection{\code{PMIx_Data_copy_payload}}
...
+
+The caller is responsible for pre-packing the provided
+payload - the load function cannot convert to network byte order
```
jjhursey, line 470:
```suggestion
payload. For example, the load function cannot convert to network byte order
```

jjhursey (commented):
Minor wording suggestions

SteVwonder:
One question from the quarterly was whether or not these functions should also be exported through the python bindings.  @jjhursey was wondering if any of the other `pmix_data_*` functions were in the python bindings.  I don't think either of those were wrapped up on the call.

jjhursey:
This PR adds some new functions. This PR is on the provisional track.

jjhursey (approved):

jjhursey:
Implementation pointers
   - Add: `PMIx_Data_load`
     - Implemented in https://github.com/openpmix/openpmix/pull/2039
   - Add: `PMIx_Data_unload`
     - Implemented in https://github.com/openpmix/openpmix/pull/2039
   - Add: `PMIx_Data_compress`
     - Implemented in https://github.com/openpmix/openpmix/pull/2068 and https://github.com/openpmix/openpmix/pull/2070
   - Add: `PMIx_Data_decompress`
     - Implemented in https://github.com/openpmix/openpmix/pull/2068 and https://github.com/openpmix/openpmix/pull/2070

raffenet (commented):
Should the source buffer arguments in these APIs include a `const` qualifier?

Chap_API_Data_Mgmt.tex
```diff
@@ -420,4 +420,201 @@ \subsection{\code{PMIx_Data_copy_payload}}
...
+
+%%%%
+\format
+
+\versionMarker{4.1}
+\cspecificstart
+\begin{codepar}
+bool
+PMIx_Data_compress(uint8_t *inbytes, size_t size,
+                   uint8_t **outbytes, size_t *nbytes);
```
raffenet, line 543:
```suggestion
PMIx_Data_compress(const uint8_t *inbytes, size_t size,
                   uint8_t **outbytes, size_t *nbytes);
```

Chap_API_Data_Mgmt.tex
```diff
@@ -420,4 +420,201 @@ \subsection{\code{PMIx_Data_copy_payload}}
...
+
+%%%%
+\format
+
+\versionMarker{4.1}
+\cspecificstart
+\begin{codepar}
+bool
+PMIx_Data_decompress(uint8_t **outbytes, size_t *nbytes,
+                     uint8_t *inbytes, size_t size);
```
raffenet, line 590:
```suggestion
PMIx_Data_decompress(uint8_t **outbytes, size_t *nbytes,
                     const uint8_t *inbytes, size_t size);
```

raffenet:
I would also advocate for consistent source and destination argument ordering for the compress/decompress APIs. For instance, zlib and libbz2 both use (dest, src) in the compress/uncompress APIs ala memcpy. Alternatively, l4z uses (src, dest) ordering. Whatever we land on, I think consistency would be good.

Chap_API_Data_Mgmt.tex
raffenet, line 590: Combined `const` and argument ordering suggestion.
```suggestion
PMIx_Data_decompress(const uint8_t *inbytes, size_t size,
                     uint8_t **outbytes, size_t *nbytes,);
```

Chap_API_Data_Mgmt.tex
```diff
@@ -420,4 +420,201 @@ \subsection{\code{PMIx_Data_copy_payload}}
...
+\end{codepar}
+\cspecificend
+
+\begin{arglist}
+\argout{outbytes}{Address where the pointer to the decompressed data region is to be returned (handle)}
+\argout{nbytes}{Address where the number of bytes in the decompressed data region is to be returned (handle)}
+\argin{inbytes}{Pointer to the source data (handle)}
+\argin{size}{Number of bytes in the source data region (\code{size_t})}
```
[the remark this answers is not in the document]
rhc54: I've never heard of `l4z`, but the arguments for these functions were designed to mirror those for `libz` as that is what backs the API in OpenPMIx. I'd stick with that ordering. If something in these definitions doesn't mirror the equivalent `libz` API, then that's an error and it should be corrected so it does so.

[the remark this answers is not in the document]
raffenet: Do you mean these functions from `zlib.h`? If so, they use consistent ordering.
```
ZEXTERN int ZEXPORT compress OF((Bytef *dest,   uLongf *destLen,
                                 const Bytef *source, uLong sourceLen));
ZEXTERN int ZEXPORT uncompress OF((Bytef *dest,   uLongf *destLen,
                                   const Bytef *source, uLong sourceLen));
```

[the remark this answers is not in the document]
rhc54: Had to look back into the code 😄  I think my comment was from an earlier time. Looking at things, we don't actually use those APIs - we use the primitives like `deflateBound` instead. So I'd say we don't really care - we can work with whatever ordering you choose.

kathrynmohror:
@rhc54 This PR was accepted as provisional on the condition that you apply the change in the comment from @bwbarrett that contains the text 'revision exception' (adding const). Please let us know if you have any questions on this.

rhc54:
@kathrynmohror I'm assuming you have a typo in your directions, and that you meant the comments from @raffenet and not @bwbarrett  - yes?

jjhursey:
 * Passed first vote with Revision Exception: 7 yes / 0 no / 1 abstain
   - Failed without the Revision Exception): 0 yes / 6 no / 2 abstain
 * I confirmed that the Revision Exception has been applied.
   - Implementation requirement completed in [this commit](https://github.com/openpmix/openpmix/pull/2196/files#diff-76796d1c2361ca1d41c4f034d1a70727305e44e365fa6aa48a558b368350dc18) in OpenPMIx
 * This has been accepted as provisional.
 * I will do the following today:
   * Update the branch
   * Merge it into `v4`
   * Cherry-pick to `master`

jjhursey (dismissed):

jjhursey (approved):

jjhursey:
* Update the branch to `v4` HEAD
 * Merge it into `v4`
 * No other text changes made

PR #333, opened 2021-01-31
Add three envars to support HWLOC shmem connections

rhc54:
Minimizing duplicate topology tree construction is therefore of great
importance. Libraries that will utilize PMIx for other purposes
can utilize the PMIx_Load_topology API to obtain a
pointer to the local topology - this helps ensure that only one topology
tree is created. However, in some cases such as lower-level libraries,
adding a dependency on PMIx is undesirable - developers of such
libraries often prefer to keep them “thin” with minimal dependencies
and as small a memory footprint as possible.

This RFC extends topology support for such cases by exposing the three HWLOC
shmem “hooks” as environmental variables:

PMIX_HWLOC_SHMEM_FILE: The backing file containing the HWLOC shared memory region.

PMIX_HWLOC_SHMEM_ADDR: The memory address of the HWLOC shared memory region.

PMIX_HWLOC_SHMEM_SIZE: The size of the HWLOC shared memory region.

### DIFF
--- a/Chap_API_Proc_Mgmt.tex
+++ b/Chap_API_Proc_Mgmt.tex
@@ -1097,14 +1097,58 @@ returning the \refconst{PMIX_ERR_NOT_SUPPORTED} error constant.
 The returned pointer may point to a shared memory region or an actual instance
 of the topology description. In either case, the description shall be treated
 as a "read-only" object - attempts to modify the object are likely to fail and
-return an error. The \ac{PMIx} library is responsible for performing any required cleanup when the client library finalizes.
+return an error. The \ac{PMIx} library is responsible for performing any required
+cleanup when the client library finalizes.
 
 \adviceuserstart
+This is the recommended method for obtaining a topology description as it
+provides a single point-of-access to the topology, regardless of how the
+topology is generated (shared memory, XML representation, or independent
+construction).
+
 It is the responsibility of the user to ensure that the \refarg{topo} argument
 is properly initialized prior to calling this \ac{API}, and to check the
 returned \refarg{source} to verify that the returned topology description is
 compatible with the user's code.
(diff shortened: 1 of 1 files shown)


rhc54:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

jjhursey:
Some related discussion links that influenced the creation of this change:
 * https://openpmix.github.io/support/faq/avoid-hwloc-dup
 * https://github.com/openpmix/openpmix/pull/2054
 * https://www.mail-archive.com/devel@lists.open-mpi.org/msg21232.html

Chap_API_Proc_Mgmt.tex
```diff
@@ -1097,15 +1097,59 @@ \subsection{\code{PMIx_Load_topology}}
...
+pointer to the local topology - this helps ensure that only one topology
+tree is created. In some cases, particularly in lower-level libraries,
+adding a dependency on \ac{PMIx} is undesirable - developers of such
+libraries often prefer to keep them “thin” with minimal dependencies
+and as small a memory footprint as possible. \ac{PMIx} provides
+additional support for such cases by exposing the three \ac{HWLOC}
+shmem “hooks” as environmental variables:
+%
+\declareEnvarNEW{PMIX_HWLOC_SHMEM_FILE}{
+The backing file containing the \ac{HWLOC} shared memory region.
```
jjhursey, line 1137:
```suggestion
The full path to the backing file containing the \ac{HWLOC} shared memory region.
```

jjhursey (commented):
Minor wording suggestion

rhc54:
Withdrawing this proposal as subsequent discussions with @bgoglin revealed problems.

PR #335, opened 2021-01-31
Provide access to resource usage for processes and nodes

rhc54:
Define two attributes by which applications and/or tools can request
resource usage of processes and nodes. Define a structure for each
case to contain the information, and associated macros for constructing
and destructing those structures.

### DIFF
--- a/Chap_API_Job_Mgmt.tex
+++ b/Chap_API_Job_Mgmt.tex
@@ -1167,3 +1167,281 @@ Log the provided information to the host environment's job record.
 }
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\section{Resource Usage}
+\label{api:resusage}
+
+Applications and tools often want to obtain a snapshot of the resources utilized by individual
+processes and/or nodes in their job. \ac{PMIx} provides attributes for querying such information
+and accompanying structures for the results.
+
+\subsection{Process statistics}
+\label{api:pstats:struct}
+
+The process statistics structure contains information available from many \acp{OS}. The field
+corresponding to any information that is not available will be set to zero.
+
+\declarestruct{pmix_proc_stats_t}
+
+\versionMarker{4.1}
+\cspecificstart
+\begin{codepar}
+typedef struct pmix_proc_stats \{
+    /* process ident info */
+    char *node;
+    pmix_proc_t proc;
+    pid_t pid;
+    char *cmd;
+    /* process stats */
+    char state[2];
+    struct timeval time;
+    float percent_cpu;
+    int32_t priority;
+    uint16_t num_threads;
+    float pss;   /* in MBytes */
+    float vsize;  /* in MBytes */
+    float rss;  /* in MBytes */
+    float peak_vsize;  /* in MBytes */
+    uint16_t processor;
+    /* time at which sample was taken */
+    struct timeval sample_time;
+\} pmix_proc_stats_t;
+\end{codepar}
+\cspecificend
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\subsubsection{Process statistics support macros}
+The following macros are provided to support the \refstruct{pmix_proc_stats_t} structure.
+
+\littleheader{Initialize the proc stats structure}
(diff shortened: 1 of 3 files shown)


rhc54:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

Chap_API_Job_Mgmt.tex
```diff
@@ -1167,3 +1167,281 @@ \subsection{Log attributes}
...
+\label{api:ndstats:attrs}
+
+The following attributes can be used to query for node statistics:
+
+\declareAttributeNEW{PMIX_NODE_RESOURCE_USAGE}{"pmix.ndstats"}{char*}{
+Return the resource usage statistics for the specified node in a \refstruct{pmix_node_stats_t} structure.
+If no node is specified, then
+results for all nodes hosting processes within the job of the requestor
+will be returned in a \refstruct{pmix_data_array_t} of \refstruct{pmix_node_stats_t} structures.
+}
```
jjhursey, line 1445: Similar to the `PMIX_PROC_RESOURCE_USAGE` should we point out that:
> This attribute can be used with qualifiers such as \refattr{PMIX_SESSION_ID} to identify the session of the namespace whose statistics are being requested.

Chap_API_Job_Mgmt.tex
```diff
@@ -1167,3 +1167,281 @@ \subsection{Log attributes}
...
+\versionMarker{4.1}
+\cspecificstart
+\begin{codepar}
+typedef struct pmix_proc_stats \{
+    /* process ident info */
+    char *node;
+    pmix_proc_t proc;
+    pid_t pid;
+    char *cmd;
+    /* process stats */
```
jjhursey, line 1194: Do we have a link to more descriptions on each of these stats? We take them from Linux so a pointer to a good manpage description would be helpful to someone new to these fields.

jjhursey (commented):
Minor wording suggestions.

jjhursey:
We should add a Python type for these.

jjhursey:
This PR adds some new structures. This PR is on the provisional track.

There are some minor notes above that need to be addressed before the presentation.

jjhursey:
We need to update the `pmix_value_t` structure with these two :
```
        pmix_proc_stats_t *pstats;
        pmix_node_stats_t *ndstats;
```

I see quite a few other fields that need to be added to the `pmix_value_t` that are in OpenPMIx v4. I'll add those in a separate PR if you can add the above two to this PR.

jjhursey:
Implementation Notes:
   - Implemented in https://github.com/openpmix/openpmix/pull/2039
   - With the exception of `PMIX_PROC_RESOURCE_USAGE` and `PMIX_NODE_RESOURCE_USAGE` which are not yet provided in OpenPMIx (easy enough to fix)

jjhursey:
Notes from 2Q 2021 ASC Meeting (day 1)
 * Comments need to be addressed. If they can be addressed before Day 2 then we can have a Revision Exception vote.
   * [this comment](https://github.com/pmix/pmix-standard/pull/335/files#r570463170) regarding the need for documentation of the various fields in the `pmix_proc_stats_t`. Should also document the fields of `pmix_node_stats_t` (what if the value is not available - should it be set to `0`?).
   * [this comment](https://github.com/pmix/pmix-standard/pull/335#issuecomment-816345439) implementation requirement for `PMIX_PROC_RESOURCE_USAGE` and `PMIX_NODE_RESOURCE_USAGE` missing in OpenPMIx.
   * [this comment](https://github.com/pmix/pmix-standard/pull/335#issuecomment-816342696) missing `pmix_proc_stats_t` and `pmix_node_stats_t` fields in the `pmix_value_t`
 * Question regarding the choice of the data type for fields in `pmix_node_stats_t` - why `float` and not `size_t`/`double`/...? Is it better to have an accessor function instead of an explicit data structure?
 * Question if there are users of this interface at the moment.

jjhursey:
 * This PR has been pushed back for the reasons summarized in [this comment](https://github.com/pmix/pmix-standard/pull/335#issuecomment-838913030)
 * Once those comments have been addressed this PR can be brought back for presentation.

jjhursey:
Triage file: https://github.com/openpmix/pmix-tests/blob/08705b12294fc5020f2af951b9bb10b36b6cec8d/check-standard/etc/openpmix_master-pmix-standard_master.txt#L277C1-L309

PR #337, opened 2021-02-26
Add Provisional Markers

jjhursey:
Add a mechanism for labeling provisional items and text segments. Also cleanup the style sheet a bit with some comments.

```
\provisionalMarker     Add a provisional label to the margin
\markProvisional       Add a provisional label to the margin and highlight the provided text

\declareconstitemProvisional  Declare Constant - Provisional
\declareEnvarProvisional      Declare Envar - Provisional
\declareAttributeProvisional  Declare Attribute - Provisional
```

### DIFF
--- a/pmix.sty
+++ b/pmix.sty
@@ -182,24 +182,78 @@
 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-% Floating version
-%\usepackage[showboxes]{textpos}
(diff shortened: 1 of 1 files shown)


jjhursey:
PR #280 will be our first Provisional item accepted. We need some LaTex markers to mark the text. Below is an example of what the markup looks like:
Screen Shot 2021-02-25 at 6 35 04 PM

abouteiller:
That's a bit too colorful to my taste, but I have nothing better to propose.

abouteiller (approved):

jjhursey:
Yeah. We can tweak the coloring if we find something that we like better. This should give us the plumbing to make that an easy change.

PR #340, opened 2021-03-02
Move from Travis to GitHub Actions

jjhursey:
* Due to rate limiting restrictions in Travis, move to GitHub Actions


jjhursey:
To see what this looks like click on the `Checks` tab in the PR below:
 * https://github.com/jjhursey/pmix-standard/pull/2

Apparently, GitHub made an agreement with DockerHub that avoids the rate limit issue. Even if they hadn't we could add an authentication token to the Action to work around it. So far this looks like it is working.

.github/workflows/build-ci.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  pull_request:
+    # We don't need this to be run on all types of PR behavior
+    # See
+    #  - https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
+    #  - https://frontside.com/blog/2020-05-26-github-actions-pull_request
+    types:
+      - opened
+      - synchronize
+      - edited
```
SteVwonder, line 12: What's the motivation for re-running on `edited`?  Based on the second link in the comment, `edited` means "title, body, or the base branch of the PR is modified".

W.r.t the defaults, this is missing `reopened`, is that intentional?

.github/workflows/build-ci.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Check out the code
+        uses: actions/checkout@v1
+      # Run the container tester
+      - name: Build and Test
+        run: docker run --rm -v ${GITHUB_WORKSPACE}:/home/pmixer/pmix-standard jjhursey/pmix-standard /bin/bash -c "cp -R /home/pmixer/pmix-standard pmix-build && cd pmix-build && make"
```
SteVwonder, line 23: I went ahead and created the `pmix` organization on docker hub (and added you @jjhursey as an owner) if we eventually want to switch to that top-level namespace:

```suggestion
        run: docker run --rm -v ${GITHUB_WORKSPACE}:/home/pmixer/pmix-standard pmix/pmix-standard /bin/bash -c "cp -R /home/pmixer/pmix-standard pmix-build && cd pmix-build && make"
```

SteVwonder (commented):
Thanks @jjhursey for putting this together!  Good to know that Github Actions are exempt from docker hub pulls.

.github/workflows/build-ci.yml
jjhursey: I included `edited` because of the base branch modification. So if they switch the PR from targeting `v4` to `master` then we should probably rerun it.

`reopened` is a good idea. I'll add that.

.github/workflows/build-ci.yml
jjhursey: Thanks for setting that up. I was thinking of moving those github repos into the pmix or openpmix GitHub namespace anyway, so it would make sense to rename the container namespace at the same time. I'll put that on my near-ish term todo list. We can make this change once that is moved over.

.github/workflows/build-ci.yml
jjhursey: Done in the latest push.

jjhursey:
@SteVwonder can you give a final review so we can merge?
I'll work on moving those containers in the next few weeks. We can update the GH Action once that's done.

SteVwonder (approved):
LGTM!

PR #347, opened 2021-04-13
Issue 175 chapter 2 master

dsolt:
This text was presented in a previous quarterly and feedback at that meeting resulted in the changes in the commit "Feedback from reading".

      we know your current thoughts

      want to be an advocate for it
      other technologies instead


dsolt:
[pmix-standard-v5.0.pdf](https://github.com/pmix/pmix-standard/files/6304816/pmix-standard-v5.0.pdf)

Chap_Terms.tex
```diff
@@ -20,15 +20,22 @@ \chapter{PMIx Terms and Conventions}
...
 
-\item \declareterm{process} refers to an operating system process, also commonly referred to as a \emph{heavyweight} process. A process is often comprised of multiple \emph{lightweight threads}, commonly known as simply \declaretermAlt{threads}{thread}.
+\item \declareterm{application} represents a set of identical, but not necessarily unique,
+execution contexts within a \emph{job}.
+
+\item \declareterm{process} is assumed for ease of presentation to be an operating system process, also commonly referred to as a \emph{heavyweight} process. A process is often comprised of multiple \emph{lightweight threads}, commonly known as simply \declaretermAlt{threads}{thread}.  However, it is not the intent of the \ac{PMIx} Standard to restrict the term process to a particular concept or implementation.
 
 \item \declaretermAlt{client}{clients} refers to a process that was registered with the \ac{PMIx} server prior to being started, and connects to that \ac{PMIx} server via \refapi{PMIx_Init} using its assigned namespace and rank with the information required to connect to that server being provided to the process at time of start of execution.
```
naughtont3, line 40: Since `client` and `server` are defined in this terms area, should there also be a `tool` definition (like what is given at the start of Ch.17 Tools and Debuggers)?  Just so major terms are included in one spot.

kathrynmohror:
@dsolt This PR passed the first vote on the condition that you apply the changes you presented as revision exception modifications.

Chap_Terms.tex
```diff
@@ -138,9 +145,8 @@ \section{Naming Conventions}
 \item The \code{pmix_} prefix is used to denote function pointer and type definitions.
 \end{itemize}
 
-Users should not use the \textbf{\code{"PMIX"}}, \textbf{\code{"PMIx"}}, or \textbf{\code{"pmix"}} prefixes in their applications or libraries so as to avoid symbol conflicts with current and later versions of the \ac{PMIx} Standard.
+Users shall not use the \textbf{\code{"PMIX_"}}, \textbf{\code{"PMIx_"}}, or \textbf{\code{"pmix_"}} prefixes for symbols in their code so as to avoid symbol conflicts with \ac{PMIx} implementations.
```
rhc54, line 149: Since you are explicitly carving out these prefixes, should you also explicitly prohibit the use of string attributes that start with "pmix"? We state this in the chapter on "reserved keys", but given that you are placing a restriction on symbols here, it might fit to repeat the attribute reservation.

jjhursey: We have a note for that a few lines up in this PR:
 * https://github.com/pmix/pmix-standard/pull/347/files#diff-cb249f8e90208e1194ad52d5e550966ceca13e78cd8512516575d7cd5bd970a4R139

> \item The string representation of attributes are prefixed with \code{"pmix"}.

jjhursey:
 * Passed first vote with Revision Exception: 7 yes / 0 no / 1 abstain
   - Failed without the Revision Exception): 1 yes / 4 no / 3 abstain
 * Eligible for second vote 3Q 2021 ASC Meeting
 * I confirmed that the Revision Exception has been applied.
 * Signed off checker failing for the "co-authored" comments. Ok to ignore for now.

Chap_Terms.tex
rhc54: Understood - but my point was that if you explicitly tell users they cannot use prefixed symbols, then you probably also need to explicitly tell them they cannot use prefixed attributes. You already cited the prefixed symbol standard above as well, so you now are implying that even though there exist other rules (e.g., the prefixed attribute), only the symbols are protected.

jjhursey:
PMIx ASC 3Q 2021
 * Passed the second vote: 8 yes / 0 no / 1 abstain
 * Will be merged into the master branch targeting PMIx v5.0

jjhursey:
I rebased and merged this PR. It'll be in the v5 version of the document.
I also fixed the signed-off error in the last commit (it doesn't handle the "co-authored" well).

jjhursey (approved):

PR #357, opened 2021-08-05
Use Case: Cross-Version Support

jjhursey:
* A use case the details the cross-version compatibility needs of a PMIx implementation.

### DIFF
--- a/App_Use_Cases.tex
+++ b/App_Use_Cases.tex
@@ -6,11 +6,11 @@
 
 The \ac{PMIx} standard provides many generic interfaces that can be composed into higher-level use cases in a variety of ways. While the specific interfaces and attributes are standardized, the use cases themselves are not (and should not) be standardized. Common use cases are included here as examples of how PMIx's generic interfaces \textit{might} be composed together for a higher-level purpose. The use cases are intended for both \ac{PMIx} interface users and library implementors. Whereby a better understanding of the general usage model within the community can help users picking up PMIx for the first and help implementors optimize their implementation for the common cases.
 
-Each use case is structured to provide background information about the high-level use case as well as specific details about how the PMIx interfaces are used within the use case. Some use cases even provide code snippets. These code snippets are apart of larger code examples located within the standard's source code repository, and each complete code example is fully compilable and runnable. The related interfaces and attributes collected at the bottom of each use case are mainly for conveinence and link to the full standardized definitions.
+Each use case is structured to provide background information about the high-level use case as well as specific details about how the PMIx interfaces are used within the use case. Some use cases even provide code snippets. These code snippets are apart of larger code examples located within the standard's source code repository, and each complete code example is fully compilable and runnable. The related interfaces and attributes collected at the bottom of each use case are mainly for convenience and link to the full standardized definitions.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-\section {Business Card Exchange for Process-to-Process Wire-up}
+\section{Business Card Exchange for Process-to-Process Wire-up}
 \label{app:uc-business-card-exchange}
 
 \subsection{Use Case Summary}
@@ -82,6 +82,8 @@ For each process this information is also useful (note that any one process may 
 
 There are other keys that are helpful to have before a synchronization point. This is not meant to be a comprehensive list.
 
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \section{Debugging}
 \label{app:uc-debugging}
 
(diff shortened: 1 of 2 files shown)


App_Use_Cases.tex
```diff
@@ -502,7 +508,85 @@ \subsection{Use Case Details}
...
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\section{Cross-Version Compatability}
+\label{app:uc-cross-version}
+
+\subsection{Use Case Summary}
+
+The \ac{PMIx} interface serves as a conduit between clients (e.g., \ac{MPI} libraries), tools (e.g., debuggers), and servers (e.g., \acp{RM}).
+As such, it is probable that a process operating in one of these roles (e.g., as a client or tool) is running a different version of the same \ac{PMIx} implementation than the process with which it is communicating that is operating in a different role (e.g., as a server).
+For processes running in containers cross-version compatibility is especially important because the container image and the system software levels will naturally evolve and drift apart.
+As such, there is a need for the \ac{PMIx} implementation to provide cross-version compatibility.
```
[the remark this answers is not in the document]
SteVwonder: There are a few more like this, but I'll avoid marking all of them until we converge on how we want to handle it.

App_Use_Cases.tex
```diff
@@ -502,7 +508,85 @@ \subsection{Use Case Details}
...
+
+The responsibility for providing cross-version compatibility is a feature of the \ac{PMIx} implementation and not necessarily of the \ac{PMIx} standard.
+The \ac{PMIx} standard must strive to enable, and never limit, both the cross-version compatibility in the \ac{PMIx} implementation, and the ability for a \ac{PMIx} consumer to adapt to cross-version differences in capabilities.
+
+This use case is focused on cross-version compatibility between versions of the same \ac{PMIx} implementation and not between different \ac{PMIx} implementations.
```
SteVwonder, line 529: Subjective nit (feel free to take or leave):
```suggestion
This use case is focused on cross-version compatibility between different versions of the same \ac{PMIx} implementation and not between different \ac{PMIx} implementations.
```

App_Use_Cases.tex
```diff
@@ -502,7 +508,85 @@ \subsection{Use Case Details}
...
+
+Cross-version compatibility responsibilities are not restricted to \ac{PMIx}, but a general issue for any library that coordinates across multiple processes. This includes, but not limited to, client/server libraries, and libraries with a user-space and kernel-space component (e.g., high-performance interconnect libraries).
```
SteVwonder, line 531: Maybe hoist this up into the opening paragraph?  Seems like this fits in well with the more general opening paragraph that sets the context.  No strong feelings though, fine if it stays here.

App_Use_Cases.tex
```diff
@@ -502,7 +508,85 @@ \subsection{Use Case Details}
...
+This handshake often occurs during initialization (though it could occur on a per-operation basis depending on the \ac{PMIx} implementation), for example during the following operations:
+\begin{itemize}
+\item PMIx Clients: \refapi{PMIx_Init}
+\item PMIx Tools: \refapi{PMIx_tool_init}, \refapi{PMIx_tool_attach_to_server}
+\item PMIx Servers: \refapi{PMIx_server_init}, \refapi{pmix_server_client_connected2_fn_t}, \refapi{pmix_server_tool_connection_fn_t}
+\end{itemize}
+
+Commonly this cross-version handshake occurs completely transparently to the consumers of the \ac{PMIx} interface since it happens inside the \ac{PMIx} implementation of these interfaces.
+However, during the negotiation, some features available in one version might not be available in the other.
+The consumer of the \ac{PMIx} interface should always be prepared to receive the \refconst{PMIX_ERR_NOT_SUPPORTED} error code from a \ac{PMIx} interface call that the other side either does not support or is not available in the version of the library with which they are linked.
```
SteVwonder, line 560: Presumably forward-compatibility here is an issue when a newer client tries to call a function/interface that doesn't exist in an older server (since the linker will fail)?  But a newer client passing in an attribute that doesn't exist in the older server would probably work and the client would (probably?) get a `PMIX_ERR_NOT_SUPPORTED`?  If so, might be good to call out that difference in the interface vs attribute forwards compatibility.

SteVwonder (commented):
Thanks @jjhursey! Just a couple of comments

App_Use_Cases.tex
rhc54: It doesn't quite work that way. The notion of cross-compatibility implies that the client is built and linked against one version of the library, and the server is built and linked against a different one. In your example, the newer client is able to build and link against the newer version of the library because the new API is defined there. When they call the new API, the library will send a request up to the server. The older version of the server library is required to return a `PMIX_ERR_NOT_SUPPORTED` value since it won't recognize the request - it specifically is _not_ allowed to segfault or silently ignore the request.

Attributes work the same way. The newer client likely won't have an issue with an older attribute, but the older server must cleanly handle any newer attribute - ignoring it if it isn't required, returning "not supported" if the attribute is marked as required.

App_Use_Cases.tex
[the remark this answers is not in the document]
jjhursey: I like the new wording. I'll make that change.

App_Use_Cases.tex
jjhursey: I'll leave it here for now and see what others think. I'm inclined to leave it where it is as a closing statement for generality.

App_Use_Cases.tex
jjhursey: I pushed a new commit with this change.

jjhursey:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

jjhursey:
PMIx ASC 4Q 2021
 * Passed the first vote: 12 yes / 0 no / 1 abstain
 * This is eligible for second vote in the 1Q 2022 ASC quarterly meeting

jjhursey:
PMIx ASC 1Q 2022

 * Passed the second vote: 11 yes / 0 no / 0 abstain
 * This is can be merged into main/v5 branches once conflicts are addressed (I'll work on that)

jjhursey:
**Editorial**: Rebased the reference branch without text changes to prepare to merge.

Issue #358, opened 2021-08-11
Constant values need to be defined


rhc54:
### Overview

The Standard currently only defines the name of PMIx constants - it does not assign them a value, thus leaving the value to each implementation. This creates a problem for vendors seeking to distribute implementation-agnostic products as they cannot know the meaning of a particular value returned by a function call.

### Motivation

Implementation-agnostic products (e.g., debuggers) are typically compiled against one implementation, but are designed to work against whatever implementation is installed on the target system. This is done by using the combination of `dlopen` and `dlsym` to load the library and access its public APIs.

Unfortunately, the value of the constants in the code is determined at compile time - it isn't possible to load them via `dlopen`. Thus, the resulting binary is "locked" to the implementation against which it was compiled, thereby forcing the vendor to distribute implementation-specific copies of their product.

### Discussion Items

The solution to the problem is to have the Standard define the actual _value_ of each constant and not just its name. Doing this forces all implementations to use the same value, thus allowing for implementation-agnostic products to be distributed.

PR #350, opened 2021-05-17
Return codes

dsolt:
A review of how return values are presented for each API to be more consistent.

PR #359, opened 2021-08-11
Assign specific values to PMIx constants

rhc54:
The Standard currently only defines the name of PMIx constants - it does
not assign them a value, thus leaving the value to each implementation.
This creates a problem for vendors seeking to distribute
implementation-agnostic products as they cannot know the meaning of a
particular value returned by a function call.

Implementation-agnostic products (e.g., debuggers) are typically compiled
against one implementation, but are designed to work against whatever
implementation is installed on the target system. This is done by using
the combination of dlopen and dlsym to load the library and access its
public APIs.

Unfortunately, the value of the constants in the code is determined at
compile time - it isn't possible to load them via dlopen. Thus, the
resulting binary is "locked" to the implementation against which it was
compiled, thereby forcing the vendor to distribute
implementation-specific copies of their product.

The solution to the problem is to have the Standard define the actual
value of each constant and not just its name. Doing this forces all
implementations to use the same value, thus allowing for
implementation-agnostic products to be distributed.

Refs https://github.com/pmix/pmix-standard/issues/358

Signed-off-by: John DelSignore <JDelSignore@perforce.com>

### DIFF
--- a/Chap_API_Event.tex
+++ b/Chap_API_Event.tex
@@ -171,7 +171,7 @@ Error in event registration.
 
 \begin{constantdesc}
 %
-\declareconstitemvalueNEW{PMIX_EVENT_SYS_BASE}{-230}
+\declareconstitemNEW{PMIX_EVENT_SYS_BASE}
 Mark the beginning of a dedicated range of constants for system event reporting.
 %
 \declareconstitemvalueNEW{PMIX_EVENT_NODE_DOWN}{-231}
@@ -180,7 +180,7 @@ A node has gone down - the identifier of the affected node will be included in t
 \declareconstitemvalueNEW{PMIX_EVENT_NODE_OFFLINE}{-232}
 A node has been marked as \emph{offline} - the identifier of the affected node will be included in the notification.
 %
-\declareconstitemvalueNEW{PMIX_EVENT_SYS_OTHER}{-330}
+\declareconstitemNEW{PMIX_EVENT_SYS_OTHER}
 Mark the end of a dedicated range of constants for system event reporting.
 %
 \end{constantdesc}
--- a/Chap_API_Event.tex
+++ b/Chap_API_Event.tex
@@ -160,7 +160,7 @@ As previously stated, upon completing its work, and prior to returning, each han
 
 \begin{constantdesc}
 %
-\declareconstitem{PMIX_ERR_EVENT_REGISTRATION}
+\declareconstitemvalue{PMIX_ERR_EVENT_REGISTRATION}{-144}
 Error in event registration.
 %
 \end{constantdesc}
@@ -171,16 +171,16 @@ Error in event registration.
 
 \begin{constantdesc}
 %
-\declareconstitemNEW{PMIX_EVENT_SYS_BASE}
+\declareconstitemvalueNEW{PMIX_EVENT_SYS_BASE}{-230}
 Mark the beginning of a dedicated range of constants for system event reporting.
 %
-\declareconstitemNEW{PMIX_EVENT_NODE_DOWN}
+\declareconstitemvalueNEW{PMIX_EVENT_NODE_DOWN}{-231}
 A node has gone down - the identifier of the affected node will be included in the notification.
 %
-\declareconstitemNEW{PMIX_EVENT_NODE_OFFLINE}
+\declareconstitemvalueNEW{PMIX_EVENT_NODE_OFFLINE}{-232}
 A node has been marked as \emph{offline} - the identifier of the affected node will be included in the notification.
 %
-\declareconstitemNEW{PMIX_EVENT_SYS_OTHER}
+\declareconstitemvalueNEW{PMIX_EVENT_SYS_OTHER}{-330}
 Mark the end of a dedicated range of constants for system event reporting.
 %
 \end{constantdesc}
@@ -557,16 +557,16 @@ The following status code may be returned indicating various actions taken by ot
 
 \begin{constantdesc}
 %
-\declareconstitem{PMIX_EVENT_NO_ACTION_TAKEN}
+\declareconstitemvalue{PMIX_EVENT_NO_ACTION_TAKEN}{-331}
 Event handler: No action taken.
 %
-\declareconstitem{PMIX_EVENT_PARTIAL_ACTION_TAKEN}
+\declareconstitemvalue{PMIX_EVENT_PARTIAL_ACTION_TAKEN}{-332}
 Event handler: Partial action taken.
 %
-\declareconstitem{PMIX_EVENT_ACTION_DEFERRED}
+\declareconstitemvalue{PMIX_EVENT_ACTION_DEFERRED}{-333}
 Event handler: Action deferred.
 %
-\declareconstitem{PMIX_EVENT_ACTION_COMPLETE}
+\declareconstitemvalue{PMIX_EVENT_ACTION_COMPLETE}{-334}
 Event handler: Action complete.
 %
 \end{constantdesc}
--- a/Chap_API_Fabric.tex
+++ b/Chap_API_Fabric.tex
@@ -81,13 +81,13 @@ The following events are defined for use in fabric-related operations.
 
 \begin{constantdesc}
 %
-\declareconstitemNEW{PMIX_FABRIC_UPDATE_PENDING}
+\declareconstitemvalueNEW{PMIX_FABRIC_UPDATE_PENDING}{-176}
 The \ac{PMIx} server library has been alerted to a change in the fabric that requires updating of one or more registered \refstruct{pmix_fabric_t} objects.
 %
-\declareconstitemNEW{PMIX_FABRIC_UPDATED}
+\declareconstitemvalueNEW{PMIX_FABRIC_UPDATED}{-175}
 The \ac{PMIx} server library has completed updating the entries of all affected \refstruct{pmix_fabric_t} objects registered with the library. Access to the entries of those objects may now resume.
 %
-\declareconstitemNEW{PMIX_FABRIC_UPDATE_ENDPOINTS}
+\declareconstitemvalueNEW{PMIX_FABRIC_UPDATE_ENDPOINTS}{-113}
 Endpoint assignments have been updated, usually in response to migration
 or restart of a process. Clients should use \refapi{PMIx_Get} to update any
 internally cached connections.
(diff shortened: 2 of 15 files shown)


rhc54:

    we know your current thoughts

    want to be an advocate for it
    other technologies instead

rhc54:
As a further note: the Storage chapter missed defining new data type values for these new data types (so we can pack/unpack them). Here is what I put into OpenPMIx to handle them:
```c
#define PMIX_STOR_MEDIUM                66
#define PMIX_STOR_ACCESS                67
#define PMIX_STOR_PERSIST               68
#define PMIX_STOR_ACCESS_TYPE           69
```
I added these here in another commit.

Chap_API_Fabric.tex
```diff
@@ -374,13 +374,13 @@ \subsection{Fabric Coordinate Views}
```
dsolt, line 374: Now that we have these definitions hard coded, do we want to get rid of the \copySignature code on line 366?  Seems unnecessary now?

Chap_API_Fabric.tex
```diff
@@ -421,10 +421,10 @@ \subsection{Fabric Operation Constants}
 
 \begin{constantdesc}
 %
-\declareconstitemNEW{PMIX_FABRIC_REQUEST_INFO}
+\declareconstitemvalueNEW{PMIX_FABRIC_REQUEST_INFO}{0}
```
dsolt, line 424: These values appear to only be used between library and RM?  If so, we could decide we don't want to specify them?

Chap_API_Server.tex
```diff
@@ -4108,10 +4108,10 @@ \subsubsection{Group Operation Constants}
 
 \begin{constantdesc}
 %
```
dsolt, line 4110: These values appear to only be used between library and RM? If so, we could decide we don't want to specify them?

Chap_API_Storage.tex
```diff
@@ -59,70 +59,70 @@ \section{Storage support constants}
...
+\declareconstitemvalueProvisional{PMIX_STORAGE_ACCESSIBILITY_REMOTE}{0x0000000000000020}
 The storage system resources are remote.
 %
 \end{constantdesc}
 
 \declarestruct{pmix_storage_persistence_t}
 \provisionalMarker{}
 
-The \refstruct{pmix_storage_persistence_t} type specifies different levels of persistence for a particular storage system.
+The \refstruct{pmix_storage_persistence_t} is a \code{uint64_t} type that defines a set of bit-mask flags for specifying different levels of persistence for a particular storage system.
```
dsolt, line 86: I think calling this out as a bitmask is fine, but if anyone is reviewing this this one is at least somewhat interesting compared to all the other constant declarations because we didn't specify this was a bitmask before and now we are.  I don't see a problem though.

Chap_API_Storage.tex
```diff
@@ -59,70 +59,70 @@ \section{Storage support constants}
...
+\declareconstitemvalueProvisional{PMIX_STORAGE_PERSISTENCE_ARCHIVE}{0x0000000000000040}
 Data on the storage system is persisted according to archive storage policies (long-term storage, typically persisted indefinitely).
 %
 \end{constantdesc}
 
 \declarestruct{pmix_storage_access_type_t}
 \provisionalMarker{}
 
-The \refstruct{pmix_storage_access_type_t} type specifies different storage system access types.
+The \refstruct{pmix_storage_access_type_t} is a \code{uint16_t} type that defines a set of bit-mask flags for specifying different storage system access types.
```
dsolt, line 116: I think calling this out as a bitmask is fine, but if anyone is reviewing this this one is at least somewhat interesting compared to all the other constant declarations because we didn't specify this was a bitmask before and now we are.  I don't see a problem though.

Chap_API_Storage.tex
```diff
@@ -23,25 +23,25 @@ \section{Storage support constants}
...
 %
-\declareconstitemProvisional{PMIX_STORAGE_MEDIUM_PMEM}
+\declareconstitemvalueProvisional{PMIX_STORAGE_MEDIUM_PMEM}{0x0000000000000020}
 The storage system uses persistent memory.
 %
-\declareconstitemProvisional{PMIX_STORAGE_MEDIUM_RAM}
+\declareconstitemvalueProvisional{PMIX_STORAGE_MEDIUM_RAM}{0x0000000000000040}
 The storage system is volatile (e.g., tmpfs).
 %
 \end{constantdesc}
```
dsolt, line 47: The advice to implementers on line 49 is probably unnecessary now.   I think it can be deleted, but maybe someone can verify.

Chap_API_Struct.tex
```diff
@@ -67,7 +67,7 @@ \section{Constants}
 
 \begin{constantdesc}
 %
-\declareconstitem{PMIX_MAX_NSLEN}
+\declareconstitemvalue{PMIX_MAX_NSLEN}{255}
 Maximum namespace string length as an integer.
 \end{constantdesc}
```
dsolt, line 73: In general, this section is going to be of more interest than the rest as far as the values of constants go.  I don't see any issues, but if we wanted to consider value changes for constants these are the ones to look at.

Chap_API_Struct.tex
```diff
@@ -882,32 +879,32 @@ \subsection{Job State Structure}
```
dsolt, line 879: What is a pmix_job_state_t used for?  I don't see anything that uses it other than a pretty print routine.

Chap_API_Struct.tex
```diff
@@ -2026,184 +2023,193 @@ \section{Generalized Data Types Used for Packing/Unpacking}
...
+\declareconstitemvalue{PMIX_INFO}{24}
 Info object.
 %
-\declareconstitem{PMIX_PDATA}
+\declareconstitemvalue{PMIX_PDATA}{25}
 Pointer to data.
 %
-\declareconstitem{PMIX_BUFFER}
-Buffer.
-%
```
dsolt, line 2109: could be a separate PR?

Chap_API_Struct.tex
```diff
@@ -2026,184 +2023,193 @@ \section{Generalized Data Types Used for Packing/Unpacking}
...
+Bitmask specifying different levels of storage accessibility (i.e,. from where a storage system may be accessed). (\refstruct{pmix_storage_accessibility_t}).
+%
+\declareconstitemvalueNEW{PMIX_STOR_PERSIST}{68}
+Bitmask specifying different levels of persistence for a particular storage system. (\refstruct{pmix_storage_persistence_t}).
+%
+\declareconstitemvalueNEW{PMIX_STOR_ACCESS_TYPE}{69}
+Bitmask specifying different storage system access types. (\refstruct{pmix_storage_access_type_t}).
+%
+\declareconstitemvalueNEW{PMIX_DATA_TYPE_MAX}{500}
 A starting point for implementer-specific data types.
```
dsolt, line 2213: PMIX_DATA_TYPE_MAX - if I had to change a value, it would be this one (making it larger).  Don't know that we need to, but it seems to most interesting one to discuss.

Chap_API_Struct.tex
```diff
@@ -2026,184 +2023,193 @@ \section{Generalized Data Types Used for Packing/Unpacking}
...
+\declareconstitemvalueNEW{PMIX_STOR_ACCESS}{67}
+Bitmask specifying different levels of storage accessibility (i.e,. from where a storage system may be accessed). (\refstruct{pmix_storage_accessibility_t}).
+%
+\declareconstitemvalueNEW{PMIX_STOR_PERSIST}{68}
+Bitmask specifying different levels of persistence for a particular storage system. (\refstruct{pmix_storage_persistence_t}).
+%
+\declareconstitemvalueNEW{PMIX_STOR_ACCESS_TYPE}{69}
+Bitmask specifying different storage system access types. (\refstruct{pmix_storage_access_type_t}).
+%
+\declareconstitemvalueNEW{PMIX_DATA_TYPE_MAX}{500}
```
dsolt, line 2212: Could be a separate PR?   Why are these added if they are not referenced anywhere?

dsolt:
Assuming we move forward with this and with governance ticket #34, then we suggest adding some overview text either to this PR or to a separate PR that explains the motivation for these changes and explains exactly what we do and do not guarantee about ABI.  The governance ticket already has some great text that we could just steal for this purpose.

Chap_API_Event.tex
```diff
@@ -168,16 +168,16 @@ \subsection{System events}
 
 \begin{constantdesc}
 %
-\declareconstitemNEW{PMIX_EVENT_SYS_BASE}
+\declareconstitemvalueNEW{PMIX_EVENT_SYS_BASE}{-230}
 Mark the beginning of a dedicated range of constants for system event reporting.
 %
```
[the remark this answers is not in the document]
rhc54: I have no rationale for assigning that specific value - the committee is welcome to change it.

Chap_API_Fabric.tex
rhc54: I think we have to retain that specification so the RM can also remain implementation agnostic (e.g., should the RM want to dlopen/dlsym the PMIx symbols)

Chap_API_Server.tex
rhc54: see above

Chap_API_Struct.tex
rhc54: Again, I have to rationale for that value - feel free to change.

Chap_API_Struct.tex
rhc54: The datatypes are defined in the Storage chapter. However, we forgot to include the datatype "type" constant for packing and unpacking purposes, so they needed to be added or else we couldn't support those types.

jdelsign:
> TV is the only one asking for this, and they have publicly stated that they have no motivation for adding PMIx support to their product.

Well, that's not completely true. Even though we have no customers that are asking for PMIx support, we'd like to add PMIx support to the TotalView debugger. However, the current lack of an implementation-agnostic ABI makes that very difficult for us. The current situation would require us to build, maintain, and distribute a separate version of the debugger for every PMIx implementation on every platform that we'd like to support. That becomes intractable very quickly.

We'd like to see the PMIx interface structured similarly to OpenMP OMPD interface. Every detail of the OMPD interface is nailed down in the OpenMP specification: the value of constants, the layout of structures, and the signatures of function calls. Further, the API provides versioning support so that the debugger and OMPD library can both support backwards compatibility. To help accomplish this, the OpenMP website provides a C-language header files that fully defines the OMPD interface. It makes it possible for TotalView to dynamically load the OMPD library and call into it without knowing anything about the implementation. It allows one distribution of TotalView on a given platform to drive multiple OMPD implementations.

rhc54:
> Well, that's not completely true. Even though we have no customers that are asking for PMIx support, we'd like to add PMIx support to the TotalView debugger. 

My apologies, John - I misunderstood. Thanks for clarifying!

jjhursey:
PMIx ASC 4Q 2021
 * Passed the first vote with revision exception: 12 yes / 0 no / 1 abstain
 * Failed the first vote without the revision exception: 6 yes / 5 no / 2 abstain
 * This is eligible for second vote in the 1Q 2022 ASC quarterly meeting

jjhursey:
PMIx ASC 1Q 2022

jjhursey:
**Editorial**: Rebased the reference branch to prepare to merge. Addressed merge conflict in `Chap_API_Struct.tex` due to the changes from the surrounding text changes from the PR:
 * https://github.com/pmix/pmix-standard/pull/350

Note that this change is currently only for master/v5.

jjhursey (approved):

PR #375, opened 2021-12-08
Add ABI language to the standard

### LINKED ISSUE #364: What does PMIx mean by "ABI"?
This question arose while working with some folks at Fedora who were packaging OpenPMIx and developed into a somewhat far-reaching discussion. They use a particular ABI checker (https://lvc.github.io/abi-compliance-checker/) to verify ABI stability between versions of a package that have the same .so number. When running it against a couple of OpenPMIx versions, the program reported some ABI "changes" that precipitated the discussion.

My response eventually satisfied the immediate question, but I feel this merits a broader discussion and eventual adoption of a "standard" definition. The ABI currently consists of several elements:

* function APIs - pretty easy to tell ABI compatibility here, and the Standard already defines these
* numerical constants - the program checks to see if any have changed in value. RFC #359 should take care of it.
* structure membership, including relative positions of elements within the structure - the Standard defines these today
* macros - their program doesn't check these, but we have discussed how to handle them
* string attributes

It is the last item that sparked the majority of the discussion and likely requires some clarification in the Standard. The ABI checker specifically took exception to the addition of several PMIx attributes. My position was that "attributes" are _not_ part of the ABI because:

(a) the Standard requires that any implementation return "not supported" if it does not recognize a required attribute, though it is allowed to ignore a non-required attribute that it does not recognize

(b) as a result of (a), an application must always be prepared to deal with the return of a "not supported" response to any PMIx function call.

Thus, two versions of a library remain ABI compatible even though new attributes may have been added, or attributes may have been removed, because passing them to a library that doesn't recognize them has a clearly defined behavior/response. An application can therefore still dynamically use either version without problem - and this is the essence of "ABI compatibility".

We should decide if the community agrees with that understanding, or wishes to adopt some other definition.

jjhursey:
* Fixes #364

### DIFF
--- a/Chap_API_Struct.tex
+++ b/Chap_API_Struct.tex
@@ -971,6 +971,11 @@ typedef struct pmix_value \{ \\
 \} pmix_value_t;
 }
 
+The size of the \refstruct{pmix_value_t} must remain constant on a given platform regardless of the compiler.
+The \refstruct{pmix_value_t} is \code{32} bytes with \code{24} bytes reserved for the \code{data} union member.
+An implementation may add a padding member to the structure to achieve the required 32-byte size.
+
+
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsubsection{Value structure support}
 The following macros and \acp{API} are provided to support the \refstruct{pmix_value_t} structure.
--- a/Chap_API_Struct.tex
+++ b/Chap_API_Struct.tex
@@ -971,11 +971,6 @@ typedef struct pmix_value \{ \\
 \} pmix_value_t;
 }
 
-The size of the \refstruct{pmix_value_t} must remain constant on a given platform regardless of the compiler.
-The \refstruct{pmix_value_t} is \code{32} bytes with \code{24} bytes reserved for the \code{data} union member.
-An implementation may add a padding member to the structure to achieve the required 32-byte size.
-
-
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsubsection{Value structure support}
 The following macros and \acp{API} are provided to support the \refstruct{pmix_value_t} structure.
--- a/Chap_Introduction.tex
+++ b/Chap_Introduction.tex
@@ -248,9 +248,9 @@ To facilitate such interoperability the PMIx \ac{ASC} maintains a set of standar
 
 In recognition that there are circumstances where the \ac{ABI} needs to be modified this section defines some guidance for making such modifications.
 Additions to the PMIx interface can occur without breaking \ac{ABI} compatibility.
-Deprecating and removing portions of the PMIx interface does break \ac{ABI} compatibility.
+Deprecating portions of the PMIx interface does not break \ac{ABI} compatibility but serves as a warning that the \ac{ABI} may be impacted in the future.
+Removing portions of the PMIx interface does break \ac{ABI} compatibility.
 Modifications to the existing PMIx interface do break \ac{ABI} compatibility.
-Any changes to the \ac{ABI} will occur in a major release and will be noted in the release notes.
 
 The PMIx \ac{ABI} is comprised of the following:
 \begin{compactitemize}
@@ -260,15 +260,19 @@ The PMIx \ac{ABI} is comprised of the following:
 \item Attributes that are required to be implemented and their string representation
 \end{compactitemize}
 
-If a data structure must be changed in a way that breaks the \ac{ABI} the authors should consider creating a new data structure if possible.
-Additionally, the authors must take care to update or duplicate existing PMIx interfaces that are impacted by the change.
+The \ac{ASC} prohibits data structure and function signature changes that break the \ac{ABI}
+(e.g., modifications to data structure member alignment and/or size).
+It is preferred that the authors create a new data structure or function and
+duplicate all impacted PMIx interfaces.
 
-Attributes that are required by the PMIx standard are a required part of the \ac{ABI}.
-Attributes that are not required by the PMIx standard may be included in the \ac{ABI} while others may not.
-Attributes that are part of the \ac{ABI} are denoted as such in their description.
-PMIx attributes that are not required by the PMIx standard have a well-defined behavior (see Section~\ref{chap:intro:attributes}) when the string representation is not recognized by the implementation.
-The caller of the interface already expects to respond to the scenario in which the implementation does not support that optional attribute.
-Excluding these optional attributes from the ABI allows an implementation to provide additional attributes that are non-standard.
+\rationalestart
+Breaking the \ac{ABI} of data structures and function signatures while using the same
+name prevents a PMIx implementation from supporting backward compatible and
+cross-version compatible \ac{PMIx} libraries. 
+\rationaleend
+
+The value associated with attributes and constants are part of an \ac{API} and not part of a \refterm{Linker ABI}.
+The PMIx Standard defines a \refterm{Build ABI} where the value associated with PMIx attributes and constants can be assumed based on the \ac{ABI} version.
 
 The PMIx Standard \ac{ABI} is defined by a set of \ac{ASC} maintained headers.
 The instantiation of the PMIx Standard \ac{ABI} in binary form is platform-specific.
@@ -287,7 +291,8 @@ Both the Stable ABI and Provisional ABI are versioned with two increasing number
 \item \texttt{MAJOR} incremented when the ABI changed in a backward-incompatible manner.
 \item \texttt{MINOR} incremented when functionality is added to the ABI in a backward-compatible manner.
 \end{compactitemize}
-
+The PMIx Standard ABI version numbers largely follow the Semantic Versioning 2.0.0 specification\footnote{Semantic Versioning \url{https://semver.org/spec/v2.0.0.html}}.
+However, a PMIx implementation may use a different version numbering technique for the objects associated with that implementation.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsubsection{Linker versus Build ABI}
--- a/Chap_Introduction.tex
+++ b/Chap_Introduction.tex
@@ -202,6 +202,7 @@ A default attribute name signifying that the attribute field of a \ac{PMIx} stru
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{PMIx Roles}
+\label{chap:intro:roles}
 
 The role of a \ac{PMIx} process in the \ac{PMIx} universe is grouped into one of three categories based on how it operates in the \ac{PMIx} environment namely as a \emph{client}, \emph{server}, or \emph{tool}.
 As a result, there are three corresponding groupings of \acp{API} each with their own initialization and finalization functions.
@@ -235,6 +236,7 @@ These \emph{client-role,server-role, and tool-role \ac{PMIx} standard compliant}
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Application Binary Interface (ABI)}
+\label{chap:intro:abi}
 
 An \acf{API} defines how data types, data structures, and functions are represented in source code.
 An \acf{ABI} defines how data types, data structures, and functions are represented in machine code for a given system making it platform-specific.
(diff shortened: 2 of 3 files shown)


jjhursey:
Leaving this in draft form while we discuss #364 - this is an attempt to start sketching the language.

Chap_Introduction.tex
```diff
@@ -233,3 +234,39 @@ \subsection{PMIx Roles}
...
+\begin{compactalphaenum}
+\item Function signatures
+\item Constants and their values
+\item Structures and their membership (including field position)
+\item Attributes that are required to be implemented and their string representation
+\end{compactalphaenum}
+
+When adding a field to an existing data structure that has padding then that padding must be adjusted so as to preserve the stability of the \ac{ABI}.
+If a data structure must be changed in a way that breaks the \ac{ABI} the authors should consider creating a new data structure if possible.
+Additionally, the authors must take care to update or duplicate existing PMIx interfaces that are impacted by the change.
```
jjhursey, line 263: Remove. No padding in the current standard. May add in the future if padding in a structure is added.

jjhursey:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

Chap_API_Struct.tex
```diff
@@ -969,6 +969,11 @@ \subsection{Value Structure}
 \} pmix_value_t;
 }
 
+The size of the \refstruct{pmix_value_t} must remain constant on a given platform regardless of the compiler.
+The \refstruct{pmix_value_t} is \code{32} bytes with \code{24} bytes reserved for the \code{data} union member.
+An implementation may add a padding member to the structure to achieve the required 32-byte size.
```
raffenet, line 974: Are we saying it should be 32 bytes regardless of platform or just on x86-64 (or ppc64le 😄 )? FWIW, if I build a 32-bit x86 executable with the current ABI headers,`pmix_value_t` is 16 bytes and the data portion is 12.

rhc54: Good point - we definitely don't want to force 32-bit builds to waste that space.

jjhursey: Good point. Do you all have a suggestion on a re-wording?

rhc54: Actually, let's take a step back. It occurs to me that we may be over complicating things here. Let's take a look at what we mean by "ABI".

I think we all agree that we should take the "build" meaning for ABI, so let's take that as a starting point. This means that a program is assured that they can access a library of the same ABI they built against, with some assurance that symbols will resolve, constants will match in value, and structs will have the same size/ordering of fields.

There is nothing in that statement to say a 64-bit program could use a 32-bit library, or vice versa. Likewise, there is no statement that indicates you can move your app across architecture without change - which is good because that is pretty much impossible anyway. If you move from x86-64 to ARM, you are going to have to recompile.

So any statements about padding should be restricted by those constraints. It should be adequate to state that the padding should align values with the native (64 or 32-bit) addressing scheme.

However, there is also nothing that proposes that a client app can interact with a server based on a different implementation. In the case of a shared memory subsystem, that would be highly unlikely to work anyway as the details of internal implementation are probably different. For non-shared memory interactions, there is no standard wire protocol - and even if there were, it wouldn't affect the ABI as both sides instantiate the objects passed over the wire themselves.

So do we even need padding? I'm thinking we don't as the only reason to pad was the shared memory subsystem, and that actually is an implementation-internal subsystem - and not an ABI. In other words, the ABI should not apply to cross-implementation interactions - it should be restricted to the interface between a program and an implementation's library.

jjhursey: The primary use case that we have for the ABI is to be able to build against the PMIx headers, then link with various PMIx implementations. This would allow a tool to build without linking to a specific PMIx implementation then use `dlopen/dlsym` to pick up the library symbols at runtime.

On a given architecture+compiler the sizeof the `pmix_value_t` should not change if the structure stays the same as it is defined in the standard. If additional fields are added to the union by an implementation beyond what is defined in the standard then it should not exceed the sizeof the data union as it exists in the standard.

Maybe that last sentence is all that we need? That will ensure that a tool built against ABI vX.Y will work with an implementation that supports ABI vX.Y even if that implementation has extra elements in the data union - since that union has not grown. So any arrays of structures that the tool has created statically are still at the correct size/offset.

I'd like to discuss this today in the call to reach a resolution.

raffenet: > Maybe that last sentence is all that we need? That will ensure that a tool built against ABI vX.Y will work with an implementation that supports ABI vX.Y even if that implementation has extra elements in the data union - since that union has not grown. So any arrays of structures that the tool has created statically are still at the correct size/offset.

IMO, yes. That should be all we need. As long as the size (and offset) of the elements in the structure are consistent, then no pad should be needed.

jjhursey: Note from April 14, 2022 meeting:
> The PMIx ABI is platform-specific. The headers are defined, but the realization of the ABI in binary form is platform-specific. Therefore care must be taken with data structures to not change the position of members, size of the structure or members in the structure, and the offset of the members in the structure for that platform.

jjhursey: I just pushed this change as a new commit.

jjhursey:
I pushed a new pair of commits addressing the comments from the discussion so far. Mirrors over the versioning text from the Governance document [here](https://github.com/pmix/governance/pull/34/files#r851349685)

Chap_Introduction.tex
```diff
@@ -232,4 +234,81 @@ \subsection{PMIx Roles}
...
+
+The PMIx \ac{ABI} is comprised of the following:
+\begin{compactitemize}
+\item Function signatures
+\item Constants and their values
+\item Structures and their membership (including field position)
+\item Attributes that are required to be implemented and their string representation
+\end{compactitemize}
+
+If a data structure must be changed in a way that breaks the \ac{ABI} the authors should consider creating a new data structure if possible.
```
abouteiller, line 263: Changing the ABI of data structures (e.g., removing/reordering existing fields) will break backward compatibility. We should add a short rationale as to why this is **very** undesirable, and not only one should *consider*, but should avoid this type of changes at all cost.

Chap_Introduction.tex
```diff
@@ -232,4 +234,81 @@ \subsection{PMIx Roles}
...
+Additionally, the authors must take care to update or duplicate existing PMIx interfaces that are impacted by the change.
+
+Attributes that are required by the PMIx standard are a required part of the \ac{ABI}.
```
abouteiller, line 266: Ralph comment: Attributes are non-linkable string objects (i.e., they are defines, they don't have an address as a variable), hence they are not an ABI element.

Chap_Introduction.tex
jjhursey:
```suggestion
If a data structure must be changed in a way that breaks the \ac{ABI} the authors are strongly advised to create a new data structure instead of breaking the \ac{ABI} and backward compatibility.
```

Let's add a clear rationale to add to this section to capture the middleware constraints dealing with cross and backwards compatibility issues.

Chap_Introduction.tex
abouteiller: ```Although the values of attributes and constants are not part of the \ac{ABI}, they are defined as standardized members of the \ac{API}, hence their value can be assumed based on the \ac{ABI} version.```

jjhursey: Please take time before Wednesday afternoon to suggest any revisions to this text.

This text will replace the full paragraph here (lines 266-271).

Chap_Introduction.tex
```diff
@@ -232,4 +234,81 @@ \subsection{PMIx Roles}
...
+The PMIx Standard ABI is defined in two parts.
+The \textit{PMIx Standard Stable ABI} represents the Stable PMIx Standard elements (see the PMIx Governance document).
+The \textit{PMIx Standard Provisional ABI} represents the Provisional PMIx Standard elements (see the PMIx Governance document).
+
+Both the Stable ABI and Provisional ABI are versioned with two increasing numbers:
+\begin{compactitemize}
+\item \texttt{MAJOR} incremented when the ABI changed in a backward-incompatible manner.
+\item \texttt{MINOR} incremented when functionality is added to the ABI in a backward-compatible manner.
+\end{compactitemize}
+
```
jjhursey, line 290: Add a comment along the lines of:
```
The PMIx Standard ABI versioning follows the Semantic versioning procedure. However, the PMIx Implementation may use a different version numbering for their library objects than the PMix Standard ABI.
```

Chap_Introduction.tex
```diff
@@ -232,4 +234,81 @@ \subsection{PMIx Roles}
...
+If a data structure must be changed in a way that breaks the \ac{ABI} the authors should consider creating a new data structure if possible.
+Additionally, the authors must take care to update or duplicate existing PMIx interfaces that are impacted by the change.
+
+Attributes that are required by the PMIx standard are a required part of the \ac{ABI}.
+Attributes that are not required by the PMIx standard may be included in the \ac{ABI} while others may not.
+Attributes that are part of the \ac{ABI} are denoted as such in their description.
+PMIx attributes that are not required by the PMIx standard have a well-defined behavior (see Section~\ref{chap:intro:attributes}) when the string representation is not recognized by the implementation.
+The caller of the interface already expects to respond to the scenario in which the implementation does not support that optional attribute.
+Excluding these optional attributes from the ABI allows an implementation to provide additional attributes that are non-standard.
```
jjhursey, line 271:
```suggestion
The value associated with attributes and constants are part of an \ac{API} and not part of a \refterm{Linker ABI}.
The PMIx Standard defines a \refterm{Build ABI} where the value associated with PMIx attributes and constants can be assumed based on the \ac{ABI} version.
```

Chap_Introduction.tex
jjhursey: I captured this change in the https://github.com/pmix/pmix-standard/pull/375/files#r870587051
Let's follow up on that comment.

Chap_Introduction.tex
jjhursey: Add the comment:
```suggestion
The PMIx Standard ABI version numbers largely follow the Semantic Versioning 2.0.0 specification\footnote{Semantic Versioning \url{https://semver.org/spec/v2.0.0.html}}.
However, a PMIx implementation may use a different version numbering technique for the objects associated with that implementation.
```

Chap_Introduction.tex
abouteiller:
```diff
+Changing data structures (in particular changes that modify the field alignment) and function signatures
+breaks the \ac{ABI}. The \ac{ASC} strongly discourages such changes. It is preferred that the authors 
+duplicate existing PMIx interfaces that are impacted by the change, 
+and potentially initiate the deprecation procedure for the duplicated interfaces.
+
+\rationalestart
+Breaking the \ac{ABI} of data structures and function signatures while using the same
+name prevents the implementation of backward compatible and cross-version compatible
+\ac{PMIx} libraries. 
+\rationaleend
```

abouteiller: I like the new text but we may need some more plenary time to discuss the implications.

jjhursey:
```suggestion
The \ac{ASC} prohibits data structure and function signature changes that break the \ac{ABI}
(e.g., modifications to data structure member alignment and/or size).
It is preferred that the authors create a new data structure or function and
duplicate all impacted PMIx interfaces.

\rationalestart
Breaking the \ac{ABI} of data structures and function signatures while using the same
name prevents a PMIx implementation from supporting backward compatible and
cross-version compatible \ac{PMIx} libraries. 
\rationaleend
```

jjhursey:
PMIx ASC 2Q 2022
 * Passed the first vote: 7 yes / 3 no / 1 abstain
 * Passed the revision exception vote: 9 yes / 3 no / 1 abstain
 * This (with revision exception) is eligible for second vote in the 3Q 2022 ASC quarterly meeting

Chap_Introduction.tex
```diff
@@ -232,4 +234,81 @@ \subsection{PMIx Roles}
...
+The PMIx \ac{ABI} is comprised of the following:
+\begin{compactitemize}
+\item Function signatures
+\item Constants and their values
+\item Structures and their membership (including field position)
+\item Attributes that are required to be implemented and their string representation
+\end{compactitemize}
+
+If a data structure must be changed in a way that breaks the \ac{ABI} the authors should consider creating a new data structure if possible.
+Additionally, the authors must take care to update or duplicate existing PMIx interfaces that are impacted by the change.
```
jjhursey, line 264:
```suggestion
```

jjhursey:
Revision exception changes have been committed to this PR as a separate commit.

jjhursey:
PMIx ASC 3Q 2022
 * Passed the second vote: 8 yes / 0 no / 0 abstain
 * This will be included in the next major release.

jjhursey:
Rebased branch. No changes made.

PR #376, opened 2021-12-24
Issue 373 - Define when PMIX_PARENT_ID is set and add definition for tool

dsolt:


rhc54:
I think the note about `PMIX_PARENT_ID` is basically correct. I can think of a way around it, but I'm not sure it is worth the bother - e.g., someone using `srun` in Slurm today technically doesn't invoke `PMIx_Spawn` (even under the covers) and yet we still set the parent ID. The method used inside Slurm is morally equivalent to calling "spawn", so I'd consider that close enough.

This question of a definition for the term "tool" continues to be problematic. It was originally a very clean/clear distinction: "clients" were given their server's contact info, while "tools" were not. The basic concept was that a "client" was a process started via a call to "spawn" (with the caveat explained above) while a "tool" was a process started at the command line. However, things expanded from that point as people started using "spawn" to start things like debuggers that are typically considered a "tool" (without anyone necessarily defining that either). We also extended access to PMIx functions such that any process that calls one of the three flavors of `PMIx_Init` has access to _all_ of the PMIx functions.

So as of this moment, the only difference between a "tool" and a "client" is that the former will search for connection info if it isn't initially given a server's URI...while a "client" will simply return an error if it isn't explicitly given that URI in its environment or in the attributes passed to `PMIx_Init`. When you get right down to it, that's a pretty thin distinction. I've already had people asking why a client can't discover connection info in a fault recovery scenario, for example, which would further shrink the differences.

I guess what I'm raising is the question of whether we should simply drop this "tool" vs "client" distinction, or perhaps define it in terms other than how they connect to a server. I'm not entirely sure of the answer. 🤷‍♂️

Chap_Terms.tex
```diff
@@ -39,6 +39,8 @@ \chapter{PMIx Terms and Conventions}
 
 \item \declaretermAlt{client}{clients} refers to a process that was registered with the \ac{PMIx} server prior to being started, and connects to that \ac{PMIx} server via \refapi{PMIx_Init} using its assigned namespace and rank with the information required to connect to that server being provided to the process at time of start of execution.
 
+\item \declaretermAlt{tool}{tools} refers to a process that may or may not have been registered with the \ac{PMIx} server prior to being started and intializes using \refapi{PMIx_tool_init}.
+
```
jjhursey, line 43: We should cross-check this with the wording in the PMIx Roles section [here](https://github.com/pmix/pmix-standard/blob/dd1cc366a34adc3c83a4c07159bc4a558f56afc3/Chap_Introduction.tex#L214-L215). Maybe keep the definition here and just reference to it in the PMIx Roles section (Similar to what we do with the `client`).

Chap_API_Reserved_Keys.tex
```diff
@@ -370,6 +370,9 @@ \subsection{Process realm attributes}
 Process identifier of the parent process of the specified process - typically
 used to identify the application process that caused the job containing the
 specified process to be spawned (e.g., the process that called \refapi{PMIx_Spawn}).
+This attribute is only provided for a process if it was created by a call to
+\refapi{PMIx_Spawn} or \refapi{PMIx_Spawn_nb}.
+client or tool.
 }
```
dsolt, line 376: "client or tool." - left over from cut-n-paste?  Remove

Chap_Terms.tex
dsolt: I have not done anything yet other than check that the definition Josh pointed out is compatible with what I've proposed here.  The definition Josh referenced doesn't say what determines if a process is tool, though it does mention that a process is a client based on calling the "client PMIx initialization routine".   It might be nice if that text had a mirror statement for tools that says a tools calls the "tool PMIx initialization routine".  Not sure how far to go with changing that text.  Would like to hear some other folk's thoughts on this.

Chap_API_Reserved_Keys.tex
```diff
@@ -370,6 +370,8 @@ \subsection{Process realm attributes}
 Process identifier of the parent process of the specified process - typically
 used to identify the application process that caused the job containing the
 specified process to be spawned (e.g., the process that called \refapi{PMIx_Spawn}).
+This attribute is only provided for a process if it was created by a call to
```
abouteiller, line 373: In most cases, launcher tools (e.g., mpiexec) will call PMIx_Spawn at the end of the launch operation chain (that may be delegated to a local daemon, e.g., prted), so it appears that in almost all cases this value will be set.

Maybe need to clarify/add an example of how to identify if the ID correspond to the launcher tool or some other type of process (e.g., application tool, client app process, etc).

jjhursey:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

Chap_API_Reserved_Keys.tex
abouteiller: ASC Feb 15, 2022 we reviewed this issue. The general idea is to do exactly what this says (that is, tools like mpiexec will also set PMIX_PARENT_ID). If this is found to cause some compatibility difficulties with existing clients that misunderstood the spec, or relied on non-compliant behavior in prior prte versions, we'll reassess need for more attributes that clarify the case dichotomy.

jjhursey:
PMIx ASC 1Q 2022

 * This is can be merged into main/v5 and v4 branches (I'll work on that)

jjhursey (approved):

PR #380, opened 2022-01-14
Add several IOF attributes

rhc54:
In the course of implementing the IO forwarding support, we identified
several attributes deemed valuable by users. These primarily control
the output format and target location (e.g., files instead of terminal),
and whether the output should be generated locally or simply relayed
to another location (e.g., a tool) for handling.

This RFC contains the definitions and indicates the APIs where those
attributes can/should be passed.

### DIFF
--- a/Chap_API_Tools.tex
+++ b/Chap_API_Tools.tex
@@ -293,9 +293,16 @@ In addition to the attributes described in \refapi{PMIx_Spawn}, the tool may opt
     \item \pasteAttributeItem{PMIX_IOF_DROP_NEWEST}
     \item \pasteAttributeItem{PMIX_IOF_BUFFERING_SIZE}
     \item \pasteAttributeItem{PMIX_IOF_BUFFERING_TIME}
+    \item \pasteAttributeItem{PMIX_IOF_OUTPUT_RAW}
     \item \pasteAttributeItem{PMIX_IOF_TAG_OUTPUT}
     \item \pasteAttributeItem{PMIX_IOF_TIMESTAMP_OUTPUT}
     \item \pasteAttributeItem{PMIX_IOF_XML_OUTPUT}
+    \item \pasteAttributeItem{PMIX_IOF_RANK_OUTPUT}
+    \item \pasteAttributeItem{PMIX_IOF_OUTPUT_TO_FILE}
+    \item \pasteAttributeItem{PMIX_IOF_OUTPUT_TO_DIRECTORY}
+    \item \pasteAttributeItem{PMIX_IOF_FILE_PATTERN}
+    \item \pasteAttributeItem{PMIX_IOF_FILE_ONLY}
+    \item \pasteAttributeItem{PMIX_IOF_MERGE_STDERR_STDOUT}
     \item \pasteAttributeItem{PMIX_NOHUP}
     \item \pasteAttributeItem{PMIX_NOTIFY_JOB_EVENTS}
     \item \pasteAttributeItem{PMIX_NOTIFY_COMPLETION}
@@ -594,6 +601,7 @@ When requesting to forward \code{stdout}/\code{stderr}, the tool can specify sev
     \item \pasteAttributeItem{PMIX_IOF_TAG_OUTPUT}
     \item \pasteAttributeItem{PMIX_IOF_TIMESTAMP_OUTPUT}
     \item \pasteAttributeItem{PMIX_IOF_XML_OUTPUT}
+    \item \pasteAttributeItem{PMIX_IOF_RANK_OUTPUT}
 \end{itemize}
 
 The \ac{PMIx} client in the tool is responsible for formatting the output stream. Note that output from multiple processes will often be interleaved due to variations in arrival time - ordering of output is not guaranteed across processes and/or nodes.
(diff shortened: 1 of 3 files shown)


rhc54:

    we know your current thoughts

    want to be an advocate for it
    other technologies instead

naughtont3:
I think the “%r” and “%n” in the PR need to be escaped to avoid having LaTeX bomb, e.g.,   \code{\%r}

naughtont3 (approved):

Chap_API_Tools.tex
```diff
@@ -724,7 +762,23 @@ \subsection{IO Forwarding attributes}
...
 Requests that the host environment intercept the specified output stream(s) and deliver it to the requesting tool instead of its current final destination. This might be used, for example, during a debugging procedure to avoid injection of debugger-related output into the application’s results file. The original output stream(s) destination is restored upon termination of the tool.
 }
-
+%
+\declareAttributeNEW{PMIX_IOF_OUTPUT_TO_FILE}{"pmix.iof.file"}{char*}{
+Direct application output into files of form "<filename>.<nspace>.<rank>.out" (for \code{stdout}) and "<filename>.<nspace>.<rank>.err" (for \code{stderr}). If \refattr{PMIX_IOF_MERGE_STDERR_STDOUT} was given, then only the \code{stdout} file will be created and both streams will be written into it.
+}
+%
+\declareAttributeNEW{PMIX_IOF_OUTPUT_TO_DIRECTORY}{"pmix.iof.dir"}{char*}{
+direct application output into files of form "<directory>/<nspace>/rank.<rank>/stdout" (for \code{stdout}) and "<directory>/<nspace>/rank.<rank>/stderr" (for \code{stderr}). If \refattr{PMIX_IOF_MERGE_STDERR_STDOUT} was given, then only the \code{stdout} file will be created and both streams will be written into it.
```
jjhursey, line 771:
```suggestion
Direct application output into files of form "<directory>/<nspace>/rank.<rank>/stdout" (for \code{stdout}) and "<directory>/<nspace>/rank.<rank>/stderr" (for \code{stderr}). If \refattr{PMIX_IOF_MERGE_STDERR_STDOUT} was given, then only the \code{stdout} file will be created and both streams will be written into it.
```

Chap_API_Tools.tex
```diff
@@ -724,7 +762,23 @@ \subsection{IO Forwarding attributes}
 \declareAttributeNEW{PMIX_IOF_REDIRECT}{"pmix.iof.redir"}{bool}{
 Requests that the host environment intercept the specified output stream(s) and deliver it to the requesting tool instead of its current final destination. This might be used, for example, during a debugging procedure to avoid injection of debugger-related output into the application’s results file. The original output stream(s) destination is restored upon termination of the tool.
 }
-
+%
+\declareAttributeNEW{PMIX_IOF_OUTPUT_TO_FILE}{"pmix.iof.file"}{char*}{
+Direct application output into files of form "<filename>.<nspace>.<rank>.out" (for \code{stdout}) and "<filename>.<nspace>.<rank>.err" (for \code{stderr}). If \refattr{PMIX_IOF_MERGE_STDERR_STDOUT} was given, then only the \code{stdout} file will be created and both streams will be written into it.
```
jjhursey, line 767: Any reason to not make the files suffixed with `.stdout` and `.stderr` to be symmetric with the `PMIX_IOF_OUTPUT_TO_DIRECTORY` directory names?

Chap_API_Tools.tex
```diff
@@ -724,7 +762,23 @@ \subsection{IO Forwarding attributes}
...
+}
+%
+\declareAttributeNEW{PMIX_IOF_OUTPUT_TO_DIRECTORY}{"pmix.iof.dir"}{char*}{
+direct application output into files of form "<directory>/<nspace>/rank.<rank>/stdout" (for \code{stdout}) and "<directory>/<nspace>/rank.<rank>/stderr" (for \code{stderr}). If \refattr{PMIX_IOF_MERGE_STDERR_STDOUT} was given, then only the \code{stdout} file will be created and both streams will be written into it.
+}
+%
+\declareAttributeNEW{PMIX_IOF_FILE_PATTERN}{"pmix.iof.fpt"}{bool}{
+Specified output file is to be treated as a pattern and not automatically annotated by nspace, rank, or other parameters. The pattern can use \code{\%n} for the namespace, and \code{\%r} for the rank wherever those quantities are to be placed. The resulting filename will be appended with ".out" for the \code{stdout} stream and ".err" for the \code{stderr} stream. If \refattr{PMIX_IOF_MERGE_STDERR_STDOUT} was given, then only the \code{stdout} file will be created and both streams will be written into it.
```
jjhursey, line 775: Same comment as above. Why not use `.stdout` and `.stderr` instead of `.out` and `.err` to be consistent with `PMIX_IOF_OUTPUT_TO_DIRECTORY`

Chap_API_Tools.tex
rhc54: Err....no? Done

Chap_API_Tools.tex
rhc54: Done

jjhursey (dismissed):

jjhursey (approved):

PR #382, opened 2022-01-15
Touch up debugger attributes and constants

rhc54:
Debuggers don't actually release processes by generating
an event - they do so by modifying a memory location. Also,
it is helpful for applications to be able to control where
they want to be stopped thru use of their own definable
breakpoints (e.g., by setting an envar). This can be made
clearer by renaming the PMIX_DEBUG_WAIT_FOR_NOTIFY attribute
to PMIX_DEBUG_STOP_IN_APP, and defining a new PMIX_BREAKPOINT
attribute by which the application can indicate where it
has internally stopped.

### DIFF
--- a/Chap_API_Tools.tex
+++ b/Chap_API_Tools.tex
@@ -310,7 +310,7 @@ In addition to the attributes described in \refapi{PMIx_Spawn}, the tool may opt
     \item \pasteAttributeItem{PMIX_LOG_COMPLETION}
     \item \pasteAttributeItem{PMIX_DEBUG_STOP_ON_EXEC}
     \item \pasteAttributeItem{PMIX_DEBUG_STOP_IN_INIT}
-    \item \pasteAttributeItem{PMIX_DEBUG_WAIT_FOR_NOTIFY}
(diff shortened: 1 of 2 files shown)


rhc54:

    we know your current thoughts

    want to be an advocate for it
    other technologies instead

jjhursey (commented):
Since this introduces a deprecation/replacement of `PMIX_DEBUG_WAITING_FOR_NOTIFY` with `PMIX_READY_FOR_DEBUG` I think we might not be able to bring it through as a provisional item but have to use the 2 vote sequence. We can talk about it at the quarterly to get a sense of the ASC.

We will have to update the Debugger Use Case as well on the main branch.

jjhursey:
PMIx ASC 2Q 2022
 * Passed the second vote: 11 yes / 0 no / 0 abstain
 * Will be merged into the next release

jjhursey:
I rebased this PR to prepare for the merge. Minor editorial changes in the Revisions change to address merge conflict with the master branch.

jjhursey (approved):

PR #384, opened 2022-01-16
Add some missing attributes and one API for server support

rhc54:
There are a few attributes servers need in order to support
singleton operations as well as user-specified time limits
on job and spawn operations. Current server module functions
do not include the ability for the server library to pass
a local status from a collective operation (i.e., when the
server library aggregates all local participation before
calling the host)

Add a few job-related error constants for clear reporting
to the user.

Add a useful variant of a data management function.


rhc54:

    we know your current thoughts

    want to be an advocate for it
    other technologies instead

Chap_API_Proc_Mgmt.tex
```diff
@@ -536,6 +554,22 @@ \subsection{Spawn attributes}
...
 Do not generate an event when this job normally terminates.
 }
+%
+\declareAttributeNEW{PMIX_ENVARS_HARVESTED}{"pmix.evar.hvstd"}{bool}{
+Environmental parameters have been harvested by the spawn requestor - the server
+does not need to harvest them.
+}
+%
+\declareAttributeNEW{PMIX_JOB_TIMEOUT}{"pmix.job.time"}{int}{
+Time in seconds before the spawned job should time out and be terminated (0 => infinite)
```
jjhursey, line 564: Is this a total runtime timeout? Kind of like the walltime limits on batch jobs?

Chap_API_Sync_Access.tex
```diff
@@ -165,6 +165,10 @@ \subsection{Fence-related attributes}
 available to each participant at the end of the operation. By default, this will include all job-level information that was locally generated by \ac{PMIx} servers unless excluded using the \refattr{PMIX_COLLECT_GENERATED_JOB_INFO} attribute.
 }
 %
+\declareAttributeNEW{PMIX_LOCAL_COLLECTIVE_STATUS}{"pmix.loc.col.st"}{pmix_status_t}{
+Status code for local collective operation being reported to the host by the server library
```
jjhursey, line 169: I had to look back at the PR comment to understand the use case for this
 > e.g., when the server library aggregates all local participation before calling the host

Can we add such a quick note example to the definition here for clarity?

So this is a status value indicating progress in the collective or a status value indicating if the collective completed successfully/erroneously? If the former, then what status values might it send up?

I'm just trying to better understand the expected usage of this attribute.

Chap_API_Proc_Mgmt.tex
rhc54: I addressed this - please see if the revised text helps.

Chap_API_Sync_Access.tex
rhc54: I addressed this - please see if the revised text helps. Short answer is that it is the latter - a single upcall reporting completion status.

Chap_API_Proc_Mgmt.tex
jjhursey: That looks good. Thanks!

Chap_API_Sync_Access.tex
jjhursey: That helps. Thanks

jjhursey:
PMIx ASC 1Q 2022

jjhursey (dismissed):

jjhursey:
**Editorial**: Rebased the reference branch without text changes to prepare to merge.

jjhursey (approved):

PR #383, opened 2022-01-16
Convert several macros to APIs

rhc54:
The PMIx convenience macros are frequently used. However, several of them
are difficult to implement as pure macros, or even as static inline functions.
In the current implementation, these were backed by public functions outside
of the Standard, thus making them less desirable.

This RFC replaces those specific macros with their functional equivalent,
thereby providing Standard APIs for these utilities. The corresponding
macros are now marked as "deprecated".

Signed-off-by: Ralph Castain <rhc@pmix.org>

PR #385, opened 2022-01-17
Provide static initializers for PMIx structs

rhc54:
Programs that utilize PMIx structures benefit from
being provided with static initializers by which global
variables can be safely declared.

### DIFF
--- a/Chap_API_Data_Mgmt.tex
+++ b/Chap_API_Data_Mgmt.tex
@@ -45,6 +45,20 @@ typedef struct pmix_data_buffer \{
 
 \ac{PMIx} provides a set of convenience macros for creating, initiating, and releasing data buffers.
 
+%%%%
+\littleheader{Static initializer for the data buffer structure}
+\declaremacro{PMIX_DATA_BUFFER_STATIC_INIT}
+
+Provide a static initializer for the \refstruct{pmix_data_buffer_t} fields.
+
+\versionMarker{4.2}
+\cspecificstart
+\begin{codepar}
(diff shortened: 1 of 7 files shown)


rhc54:

    we know your current thoughts

    want to be an advocate for it
    other technologies instead

Chap_API_Data_Mgmt.tex
```diff
@@ -45,6 +45,20 @@ \section{Support Macros}
...
+%%%%
+\littleheader{Static initializer for the data buffer structure}
+\declaremacro{PMIX_DATA_BUFFER_STATIC_INIT}
+
+Provide a static initializer for the \refstruct{pmix_data_buffer_t} fields.
+
+\versionMarker{4.2}
+\cspecificstart
+\begin{codepar}
+PMIX_DATA_BUFFER_STATIC_INIT
```
jjhursey, line 57: What arguments does this macro take? Should we make it a function per PR #383?

jjhursey (commented):
What arguments do these macro take?
Should we make them functions per PR #383?

Chap_API_Data_Mgmt.tex
jjhursey: Thanks. So similar to `PTHREAD_MUTEX_INITIALIZER` it is the right-hand side of the initialization for a variable declaration.
 * Ref: https://linux.die.net/man/3/pthread_mutex_init

jjhursey:
PMIx ASC 1Q 2022

jjhursey (approved):

PR #387, opened 2022-01-19
Issue 175 chapter 5 8 rebase

dsolt:
This very large change set covers a thorough review of chapters 5-8 of the PMIx standard.  The resulting text actually creates new chapters, renames chapters and moves content between chapters.  Our intention is to divide this PR up into at least 3 separate PR's that will be easier to digest.

### DIFF
--- a/Chap_API_Reserved_Keys.tex
+++ b/Chap_API_Reserved_Keys.tex
@@ -14,7 +14,7 @@ posting reserved keys using the \refapi{PMIx_Put} \ac{API}.
 \ac{PMIx} implementations may choose to define their own
 custom-prefixed keys which may adhere to either the \emph{reserved} or the \emph{non-reserved} retrieval rules at the discretion of the implementation. Implementations may choose to provide such custom keys at client start of execution, but this is not required.
 
-Host environments may also opt to define their own custom keys. However, \ac{PMIx} implementations are unlikely to recognize such host-defined keys and will therefore treat them according to the \emph{non-reserved} rules described in Chapter \ref{chap:nrkeys}. Users
+Host environments may also opt to define their own custom keys. However, \ac{PMIx} implementations are unlikely to recognize such host-defined keys and will therefore treat them according to the \emph{non-reserved} rules described in Section \ref{chap:data_sharing:non_rsvd_keys}. Users
 are advised to check both the local \ac{PMIx} implementation and host environment documentation
 for a list of any custom prefixes they must avoid, and to learn of any non-standard keys that may require special handling.
 
@@ -193,7 +193,7 @@ Process map equivalent to \refattr{PMIX_PROC_MAP} expressed in Argonne National 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Job realm attributes}
-\label{chap:res:jrealm}
+\label{chap:api_rsvd_keys:jrealm}
 
 Job-related information is retrieved by including the namespace of the target
 job and a rank of \refconst{PMIX_RANK_WILDCARD} in the \refarg{proc} argument
@@ -270,7 +270,7 @@ Number of applications in the specified job.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Application realm attributes}
-\label{chap:res:aprealm}
+\label{chap:api_rsvd_keys:aprealm}
 
 Application-related information can only be retrieved by including the
 \refattr{PMIX_APP_INFO} attribute in the \refarg{info} array passed to
@@ -334,7 +334,7 @@ Regular expression describing the result of the process mapping.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Process realm attributes}
-\label{chap:res:prealm}
+\label{chap:api_rsvd_keys:prealm}
 
 Process-related information is retrieved by referencing the namespace
 and rank of the target process in the call to \refapi{PMIx_Get}. If information
@@ -431,7 +431,7 @@ In addition, process-level information includes functional attributes directly a
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Node realm keys}
-\label{chap:res:nrealm}
+\label{chap:api_rsvd_keys:nrealm}
 
 Information regarding the local node can be retrieved by directly requesting the node realm key in the call to \refapi{PMIx_Get} - the keys for node-related information are not shared across other realms. The target process identifier will be ignored for keys that are not dependent upon it. Information about a node other than the local node can be retrieved by specifying the
 \refattr{PMIX_NODE_INFO} attribute in the \refarg{info} array along with
@@ -493,7 +493,7 @@ In addition, node-level information includes functional attributes directly asso
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \section{Retrieval rules for reserved keys}
-\label{chap:rkeys:retrules}
+\label{chap:api_rsvd_keys:retrules}
 
 The retrieval rules for reserved keys are relatively simple as the keys are
 required, by definition, to be available when the client begins execution.
@@ -535,7 +535,7 @@ namespace, then the server shall take one of the following actions:
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Accessing information: examples}
-\label{chap:api_kv:getex}
+\label{chap:api_rsvd_keys:getex}
 
 This section provides examples illustrating methods for accessing information from the various realms. The intent of the examples is not to provide comprehensive coding guidance, but rather to further illustrate the use of \refapi{PMIx_Get} for obtaining information on a \refterm{session}, \refterm{job}, \refterm{application}, \refterm{process}, and \refterm{node}.
 
@@ -592,7 +592,7 @@ rc = PMIx_Get(NULL, PMIX_NUM_NODES, info, 2, &value);
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsubsection{Job-level information}
 
-Information regarding a job can be obtained by the methods detailed in Section \ref{chap:res:jrealm}. Example requests are shown below:
+Information regarding a job can be obtained by the methods detailed in Section \ref{chap:api_rsvd_keys:jrealm}. Example requests are shown below:
 
 \cspecificstart
 \begin{codepar}
@@ -618,7 +618,7 @@ rc = PMIx_Get(&wildcard, PMIX_NUM_NODES, &info, 1, &value);
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsubsection{Application-level information}
 
-Information regarding an application can be obtained by the methods described in Section \ref{chap:res:aprealm}. Example requests are shown below:
+Information regarding an application can be obtained by the methods described in Section \ref{chap:api_rsvd_keys:aprealm}. Example requests are shown below:
 
 \cspecificstart
 \begin{codepar}
@@ -664,13 +664,13 @@ rc = PMIx_Get(NULL, PMIX_NUM_NODES, appinfo, 2, &value);
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsubsection{Process-level information}
 
-Process-level information is accessed by providing the namespace and rank of the target process. In the absence of any directive as to the level of information being requested, the \ac{PMIx} library will always return the process-level value. See Section \ref{chap:res:prealm} for details.
+Process-level information is accessed by providing the namespace and rank of the target process. In the absence of any directive as to the level of information being requested, the \ac{PMIx} library will always return the process-level value. See Section \ref{chap:api_rsvd_keys:prealm} for details.
 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsubsection{Node-level information}
 
-Information regarding a node within the system can be obtained by the methods described in Section \ref{chap:res:nrealm}. Example requests are shown below:
+Information regarding a node within the system can be obtained by the methods described in Section \ref{chap:api_rsvd_keys:nrealm}. Example requests are shown below:
 
 \cspecificstart
 \begin{codepar}
--- a/Chap_API_Reserved_Keys.tex
+++ b/Chap_API_Reserved_Keys.tex
@@ -9,14 +9,13 @@
 \code{"pmix"}. By definition, reserved keys are provided by the host
 environment and the \ac{PMIx} server, and are required to be available at client
 start of execution. \ac{PMIx} clients and tools are therefore prohibited from
-posting reserved keys using the \refapi{PMIx_Put} \ac{API}.
+posting reserved keys.
 
-\ac{PMIx} implementations may choose to define their own
-custom-prefixed keys which may adhere to either the \emph{reserved} or the \emph{non-reserved} retrieval rules at the discretion of the implementation. Implementations may choose to provide such custom keys at client start of execution, but this is not required.
-
-Host environments may also opt to define their own custom keys. However, \ac{PMIx} implementations are unlikely to recognize such host-defined keys and will therefore treat them according to the \emph{non-reserved} rules described in Section \ref{chap:data_sharing:non_rsvd_keys}. Users
+Host environments may opt to define non-standardized reserved keys. 
+All reserved keys, whether standardized or non-standardized, follow the same retrieval rules.
+Users
 are advised to check both the local \ac{PMIx} implementation and host environment documentation
-for a list of any custom prefixes they must avoid, and to learn of any non-standard keys that may require special handling.
+for a list of any non-standardized reserved keys they must avoid, and to learn of any non-standard keys that may require special handling.
 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
--- a/Chap_API_Reserved_Keys.tex
+++ b/Chap_API_Reserved_Keys.tex
@@ -82,8 +82,9 @@ This is required as many environments only guarantee unique namespaces within a
 session, and not across sessions.
 \adviceuserend
 
-The \ac{PMIx} server has corresponding attributes the host can use to specify the realm of information that it provides during namespace registration (see Section \ref{chap:api_server:assemble}).
-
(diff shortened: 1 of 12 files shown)


Chap_API_Reserved_Keys.tex
```diff
@@ -9,14 +9,16 @@ \chapter{Reserved Keys}
...
-posting reserved keys using the \refapi{PMIx_Put} \ac{API}.
+posting reserved keys.
 
-\ac{PMIx} implementations may choose to define their own
-custom-prefixed keys which may adhere to either the \emph{reserved} or the \emph{non-reserved} retrieval rules at the discretion of the implementation. Implementations may choose to provide such custom keys at client start of execution, but this is not required.
+%% TODO: Should we note that although values are required to be available at
+%% client start of execution, that some values can change during execution?
 
-Host environments may also opt to define their own custom keys. However, \ac{PMIx} implementations are unlikely to recognize such host-defined keys and will therefore treat them according to the \emph{non-reserved} rules described in Chapter \ref{chap:nrkeys}. Users
+Host environments may opt to define non-standardized reserved keys.
```
rhc54, line 17: Are you saying a host environment is allowed to define a key that violates the "pmix" prefix rule - i.e., it starts with "pmix"?? This seems like a risky decision as there is no way to guarantee it won't conflict with something in an implementation. I would strongly advise against it.

Chap_API_Reserved_Keys.tex
```diff
@@ -118,7 +118,8 @@ \subsection{Session realm attributes}
 }
 %
 \declareAttribute{PMIX_TDIR_RMCLEAN}{"pmix.tdir.rmclean"}{bool}{
-Resource Manager will cleanup assigned temporary directory trees.
+The Resource Manager will remove any directories or files it creates in
+\refattr{PMIX_TMPDIR}.
```
rhc54, line 122: Not quite correct, I believe. It means the RM will cleanup all the dirs and/or files that are placed in the PMIx session directories. In other words, the RM is going to cleanup the PMIx temp directory tree (i.e., the files _PMIx_ creates, not the RM), so PMIx itself doesn't need to worry about it.

Chap_API_Reserved_Keys.tex
```diff
@@ -193,32 +193,38 @@ \subsection{Session realm attributes}
...
 identifying the desired target session. This is required as
 many environments only guarantee unique namespaces within a session, and not
 across sessions.
 
-Job-level information includes the following attributes:
+The following keys, by default, request job-level information:
+They will return information about the job indicated in \refarg{proc}:
 
-%
-\declareAttribute{PMIX_NSPACE}{"pmix.nspace"}{char*}{
```
rhc54, line 212: Did this get moved somewhere else? It is required and used in various places - e.g., to retrieve my namespace when someone doesn't want to globally store the `pmix_proc_t` returned by `PMIx_Init`

Chap_API_Reserved_Keys.tex
```diff
@@ -267,32 +273,48 @@ \subsection{Job realm attributes}
...
-of the specified process is \refconst{PMIX_RANK_WILDCARD},
+shall default to the application containing the process specified by \refarg{proc}. 
+If the rank specified in \refarg{proc} is \refconst{PMIX_RANK_WILDCARD},
 then the application number shall default to that of the calling process if the
 namespace is its own job, or a value of zero if the namespace is that of a
 different job.
 
-Application-level information includes the following attributes:
+The following keys, by default, request application-level information:
+They will return information about the application indicated in \refarg{proc}:
```
rhc54, line 312: Two colons on successive sentences

Chap_API_Reserved_Keys.tex
rhc54: I see this was moved to the process realm - no issues with doing so as long as it is somewhere.

Chap_API_Reserved_Keys.tex
```diff
@@ -431,14 +479,22 @@ \subsection{Process realm attributes}
...
+These special keys can only target the local node and also require that a namespace 
+be specified using the \refarg{proc} argument to \refapi{PMIx_Get}.
 
-Information regarding the local node can be retrieved by directly requesting the node realm key in the call to \refapi{PMIx_Get} - the keys for node-related information are not shared across other realms. The target process identifier will be ignored for keys that are not dependent upon it. Information about a node other than the local node can be retrieved by specifying the
-\refattr{PMIX_NODE_INFO} attribute in the \refarg{info} array along with
-either the \refattr{PMIX_HOSTNAME} or \refattr{PMIX_NODEID} qualifiers for
-the node of interest.
 
-Node-level information includes the following keys:
+The following keys, by default, request node-level information:
```
rhc54, line 496: Two colons on successive lines

Chap_API_Reserved_Keys.tex
```diff
@@ -431,14 +479,22 @@ \subsection{Process realm attributes}
...
 \subsection{Node realm keys}
-\label{chap:res:nrealm}
+\label{chap:api_rsvd_keys:nrealm}
+
+Node-related information can be retrieved by requesting a key which defaults
+to the node realm or by including the \refattr{PMIX_NODE_INFO} attribute
+in the \refarg{info} array passed to \refapi{PMIx_Get}.
+The target node defaults to the local node unless a different node is specified
+in the \refarg{info} array using
+either the \refattr{PMIX_HOSTNAME} or \refattr{PMIX_NODEID}.
```
rhc54, line 489: We also do support requesting node level info for the node where the provide procID is executing. So it really is that you can specify the node (using hostname or nodeID), or request a node-realm key and specify a procID for the proc executing on the target node.

Chap_API_Reserved_Keys.tex
```diff
@@ -461,41 +517,24 @@ \subsection{Node realm keys}
...
-%
-\declareAttribute{PMIX_LOCAL_CPUSETS}{"pmix.lcpus"}{pmix_data_array_t}{
-A \refstruct{pmix_data_array_t} array of string representations of the \ac{PU} binding bitmaps applied to each local \refterm{peer} on the caller's node upon launch. Each string shall begin with the name of the library that generated it (e.g., "hwloc") followed by a colon and the bitmap string itself. The array shall be in the same order as the processes returned by \refattr{PMIX_LOCAL_PEERS} for that namespace.
-}
-%
-\declareAttribute{PMIX_LOCAL_SIZE}{"pmix.local.size"}{uint32_t}{
-Number of processes in the specified job or application realm on the caller's node. Defaults to job realm unless the \refattr{PMIX_APP_INFO} and the \refattr{PMIX_APPNUM} qualifiers are given.
-}
 
+SOLT2
```
rhc54, line 526: Err...what does "SOLT2" stand for?

Chap_API_Reserved_Keys.tex
```diff
@@ -461,41 +517,24 @@ \subsection{Node realm keys}
 Total available physical memory on a node.
 }
 
-\vspace{\baselineskip}
-
-The following attributes only return information regarding the \emph{caller's} node - any node-related qualifiers shall be ignored. In addition, these attributes require specification of the namespace in the target process identifier except where noted - the value of the rank is ignored in all cases.
```
rhc54, line 466: Errr...where did all these go? The comment that was removed is actually outdated anyway. You can retrieve the `PMIX_LOCAL_PEERS`, `PMIX_LOCAL_SIZE`, `PMIX_LOCAL_PROCS`, `PMIX_LOCAL_CPUSETS`, and `PMIX_LOCALLDR` for any node. PRRTE routinely provides them.

Chap_API_Sharing_Basics.tex
```diff
@@ -0,0 +1,630 @@
...
+The string representation and the macro are equivalent methods to refer to a key.
+Throughout this document, the macro is used rather than its string value to refer to reserved keys.
+Clients cannot create reserved keys, nor can they modify the value of reserved keys.
+The way reserved keys are accessed by clients 
+is different than for non-reserved keys as explained in \ref{chap:api_rsvd_keys}.
+
+Host environments are not required to provide all keys that are documented in the standard.  
+However, when providing keys that are documented, they must adhere to the behavior defined by the standard for reserved keys.  
+An implementation cannot provide a reserved key that behaves differently than what is in the standard.  
+Implementations are encouraged to standardize any non-standardized, reserved keys that they provide to avoid conflicting with other implementations or efforts to standardize the same key.
```
rhc54, line 47: It is implied here, but you might want to explicitly state that an implementation can define reserved keys outside of the standard. However, host environments are not permitted to do so as they can conflict with implementations.

Chap_API_Sharing_Basics.tex
```diff
@@ -0,0 +1,630 @@
...
+
+Host environments are permitted to provide non-reserved keys and non-standardized reserved keys.  
+However there are important caveats to providing either of these categories of keys
```
rhc54, line 50: Err - I don't think this is a good idea. High potential for conflicts between the host and implementation on reserved keys. Hosts can certainly provide non-reserved keys if they choose - advise them to prefix accordingly (e.g., `LSF_FOO` and `lsf.foo`).

Chap_API_Sharing_Basics.tex
```diff
@@ -0,0 +1,630 @@
...
+After initialization, the client can access reserved keys using the PMIx_Get API.  
+Common information that a client may access includes, but is not limited to,
+the name of the host it is executing on, 
+the number of peer processes in its namespace,
+the number of peer processes in its application,
+and its job rank.
+
+%%% (Do we want to make any other statements about how and when values can change during 
+%%%  the course of execution.
+%%%  PMIX_EXIT_CODE for example, UNIVERSE_SIZE could change if more resources are added?)
```
rhc54, line 71: I don't think we have any idea how to do this just yet, so I wouldn't make statements about it in the standard.

Chap_API_Sharing_Basics.tex
```diff
@@ -0,0 +1,630 @@
...
+Non-reserved keys are provided by clients.
+A call to PMIx_Put will result in a key-value being associated with the calling client within 
+its process data realm.
+There is no mechanism to retract a key-value that has been made available by a client through \refapi{PMIx_Put}.
+The key can be assigned a new value, but cannot be removed entirely.
+Although key values can originate from different sources, they are always retrieved using the
+\refapi{PMIx_Get} and \refapi{PMIx_Get_nb} API.
+
+The publish/lookup data realm is accessed through a separate set of API's.
+\refapi{PMIx_Put} cannot add or modify key-values within the publish/lookup realm and
```
rhc54, line 82: `PMIx_Put` has nothing to do with publish/lookup - I find this confusing mainly because of reuse of the "realm" term. You might want to reserve "realm" for the session/job/app/proc discussion to avoid confusion.

Chap_API_Sharing_Basics.tex
```diff
@@ -0,0 +1,630 @@
...
+There is no mechanism to retract a key-value that has been made available by a client through \refapi{PMIx_Put}.
+The key can be assigned a new value, but cannot be removed entirely.
+Although key values can originate from different sources, they are always retrieved using the
+\refapi{PMIx_Get} and \refapi{PMIx_Get_nb} API.
+
+The publish/lookup data realm is accessed through a separate set of API's.
+\refapi{PMIx_Put} cannot add or modify key-values within the publish/lookup realm and 
+\refapi{PMIx_Lookup} cannot access key-values outside the publish/lookup realm.
+This data realm is described in detail is chapter [Reference to publish/lookup chapter].
+Although PMIx_Publish and PMIx_Lookup are analogous to PMIx_Put and PMIx_Get in that both pairs
```
rhc54, line 85: Why not just leave publish/lookup to the chapter on that subject? It gets confusing here as it starts mixing pub/lookup with put/get, and those are fundamentally very different.

Chap_API_Sharing_Basics.tex
```diff
@@ -0,0 +1,630 @@
...
+These interactions occur within the process data realm.
+The 'put/commit/get' exchange pattern is often used to exchange process related information.
+Optionally, a 'put/commit/fence/get' exchange pattern adds the 'fence' synchronization 
+(and possible collective exchange) for applications that desire it.
+Commonly, these exchange patterns are used in a 
+\declareterm{business card exchange} (a.k.a. \declareterm{modex exchange}) 
+where one \ac{PMIx} client shares its connectivity information, then other \ac{PMIx} clients access that information to establish a connection with that client.
+In some environments that support ``instant-on'' all connectivity information for \ac{PMIx} clients is stored in the job-level information at process creation time and is accessible to the clients without the need to perform any additional key-value exchange.
+
+Keys can exist in multiple data realms, possibly with different values.
```
rhc54, line 108: This feels redundant to prior sections. Perhaps I'm missing the point of this chapter - how is what is presented different from what is in the reserved and non-reserved sections?

Chap_API_Reserved_Keys.tex
dsolt: We thought that we were simplifying the existing text.  It was our understanding that "custom keys" was referring to "non-standardized reserved keys".   We took it to mean that because this whole selection is talking about reserved keys.    I guess you are saying that the old text assumes that jsm_foo or srun_bar are custom key which are not reserved (the definition of reserved is clear... starts with "pmix_") but they are made available at the start of execution and cannot be set or changed by the client?  We assumed that custom keys would be something like "pmix_jsm_foo" or "pmix_srun_bar", though I can see now that the text about "for a list of any custom prefixes" gave a hint to what was being said that we missed.  

Assuming my above explanation is accurate, then we have a few choices... 1) We can revert back to the old text 2) We can explain the old text better to avoid the confusion we experienced in thinking a custom-key was a non-standardized reserved key 3) We can move in the direction of the simplified explanation we gave possibly adding some a warning that using non-standarized reserved keys is dangerous because it may conflict with future standardization.

rhc54: Yeah, it probably needs clarification. "Custom keys" cannot start with "pmix". The point I was trying to make is that PMIx itself won't understand them, so you cannot use them to direct PMIx behavior. Pretty much like any non-reserved key, in hindsight, so maybe just saying that "custom keys must be non-reserved" (in better wording) would suffice.

dsolt: I'll just add that, the new text has the problem that non-standardized keys beginning with pmix_ can be used by implementations and 1) that could be confusing for user's who would expect them to be able to look them up in the standard 2) they might get defined by the standard later to behave differently than the implementation originally used them.   
The old text has the problem that 1) It attempts to define this class of keys that are neither reserved nor non-reserved.   They don't follow reserved retrieval rules, but they shouldn't be set by the user either.  It's really a not well defined set of keys 2)  User's have to constantly be on the look for conflicts with implementations.  Their old code may not run on a new version of the host environment because the host environment added a custom key that conflicted with the client's use of a non-reserved key.

Chap_API_Sharing_Basics.tex
dsolt: I have some cleanup to do and we are going to divide this up into 3 PRs that will make this clearer.  The file API_nonreserved_keys.txt is going away (or being renamed to Chap_API_Sharing_Basics.tex.  So this is only presented once.  When I repost this I will use the correct git mv to move it and make it clearer.

Chap_API_Sharing_Basics.tex
dsolt: We have talked about this.  I tried to extend the concept of data realms to all keys.  At the time, its seemed clearer to say that all keys belong to a realm and realms do not overlap.   This seems easier to present rather than saying that "keys that are accessed via Get or Query are in realms, but keys accessed by publish/lookup are not in a realm".  However, it does have the undesirable side effect that you point out that you are sometimes forced to talk about publish/lookup at times when it would be more convenient to not address those functions.   I can go either way, it's a pretty minor change to just make realms about non-published data.

Chap_API_Sharing_Basics.tex
dsolt: I hear you.  See my previous comment.  I'm ok with what your saying.  Anyone else have thoughts?

Chap_API_Reserved_Keys.tex
rhc54: Perhaps the best way to describe this is:

- host environments, just like applications, may define their own non-reserved keys
- such keys must obey the non-reserved key rules regarding prefixing, and will be retrieved based on the rules for non-reserved keys
- host environments and applications are strongly advised to use appropriate prefixing on any non-reserved keys they define. You can even provide some guidance - e.g., it wouldn't be very smart for LSF to define keys starting with "ompi". It would, however, make sense for them to define keys starting with "lsf".

Users don't have to worry about conflicts with implementations as any PMIx implementation should be sticking with the reserved keys. You do indeed have to worry about conflicts with the host environment - but that is why users should be sensibly prefixing their keys. For example, you might expect an app named "foo" to prefix any key they define with "foo". Or to do like MPICH does and stick process IDs at the beginning.

As for an app not defining a value for a host environment key - I'm not sure we can/should say anything about it. Some environments might well define a key for apps to use when, for example, communicating a request to the RM. I don't see why PMIx cares either way. If the environment prohibits it, then they (a) should say so and (b) protect themselves against the inevitable violation of that prohibition.

Chap_API_Sharing_Basics.tex
dsolt: You said earlier that you thought it was a bad idea that we allow implementations to define reserved keys.  Do you really mean reserved keys here or do you mean keys that begin with something other than "pmix_", but are made available at the start of client execution?   Right now (both in current standard and this proposed change), reserved key means any key that starts with "pmix_".

Chap_API_Reserved_Keys.tex
dsolt: I just put that in there to see if anyone actually reviewed our proposed changes :)

Chap_API_Reserved_Keys.tex
dsolt: We were going for something more definitive than "cleanup", but admit we were not sure what exactly happens.   Would this be better (or can you suggest what would be most accurate?):

rhc54: So what happens is this: the host environment passes down the `PMIX_TMPDIR` which gives the PMIx server a root directory that is known "safe". The server then puts things like its shared memory segment and rendezvous files under that location. When the server finalizes, it normally will delete everything it put under `PMIX_TMPDIR`.

What this attribute declares is that the RM (i.e., the host) will delete this for us, usually by deleting the entire `PMIX_TMPDIR` directory tree. Thus, the PMIx server doesn't bother to delete anything - it just finalizes, leaving whatever files were created there.

Chap_API_Sharing_Basics.tex
rhc54: I'm not wild about implementations doing it. However, the sentence implies that it is okay - we just "encourage" them to file an RFC about them. Obviously, we can't stop anyone from defining such a key - heck, I do it in OpenPMIx just to move things along. So maybe this is the best we can do? My main concern really was the line that followed and to ensure that hosts do not define reserved keys.

jjhursey:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

dsolt:
I am working on dividing this PR up into several smaller PR's, but will leave this open until that is done and until the comments here are all resolved.

jjhursey:
This PR was broken into 4 other PRs to improve our ability to review. See tickets listed below:
   - https://github.com/pmix/pmix-standard/pull/390
   - https://github.com/pmix/pmix-standard/pull/391
   - https://github.com/pmix/pmix-standard/pull/392
   - https://github.com/pmix/pmix-standard/pull/393

Chap_API_Reserved_Keys.tex
dsolt: So should we say, "The Resource Manager will remove the directory \refattr{PMIX_TMPDIR} at server finalization time"?

rhc54: Well...not exactly. The RM will remove the directory whenever it decides to do it. It probably would do it once it finalizes the PMIx server library - but that's totally at their discretion. I would just leave it as "the RM will remove the directory".

PR #389, opened 2022-01-25
Move synchronization function to its own chapter

dsolt:
We moved the synchronization material out of Sync_Access into its own
chapter. The new Chap_API_Sharing_Basics.tex is identical to the old
Chap_API_Sync_Access chapter except the introduction changed and the
synchronization API's are deleted.  The Sharing_Basics introduction
is very minimal right now, but we have additional changes coming to
populate this chapter with an thorough explanation of the get/put
routines.

Signed-off-by: dsolt@us.ibm.com


dsolt:
Closing in favor of a new PR that uses 2 commits so you can see more clearly what was changed.  I could have just force-pushed to this branch, but  I wanted to keep this in case we want to go back to it.

Issue #365, opened 2021-10-26
Defining a PMIx ABI


rhc54:
### Overview

The community believes that it would benefit users of PMIx if we define a standard ABI that compliant implementations must provide and ensure remains stable across versions.

### Motivation

Implementation-agnostic products (e.g., debuggers) are typically compiled against one implementation, but are designed to work against whatever implementation is installed on the target system. This is done by using the combination of `dlopen` and `dlsym `to load the library and access its public APIs.

Key to utilizing this approach is the definition of a standardized ABI that is well-characterized and stable between both implementations and releases of each implementation. This issue has been created to identify the steps required to accomplish that goal, and to track progress in achieving it.

### Stages

- [x] Define what "ABI" means in the context of PMIx (see #364)
- [x] Define fixed values for PMIx error constants (see #359)
- [x] Define fixed values for non-error constants like `PMIX_ALLOC_NEW` (see #359)
- [x] Decide on values for PMIx constant boundaries (e.g., `PMIX_DATA_TYPE_MAX`) (see #394)
        - PMIX_EVENT_SYS_BASE
        - PMIX_EVENT_SYS_OTHER
        - PMIX_PROC_STATE_UNTERMINATED
        - PMIX_PROC_STATE_TERMINATED
        - PMIX_PROC_STATE_ERROR
        - PMIX_JOB_STATE_UNTERMINATED
        - PMIX_JOB_STATE_TERMINATED
        - PMIX_JOB_STATE_TERMINATED_WITH_ERROR
        - PMIX_EXTERNAL_ERR_BASE
        - PMIX_DATA_TYPE_MAX
        - PMIX_ALLOC_EXTERNAL
- [x] Review/revise the `pmix_proc_t` definition to be more memory footprint friendly (#366)
   - Revise PMIX_MAX_NSLEN?
   - Convert `pmix_nspace_t` type to a `char*` pointer?
   - No changes to these structures is being proposed.
- [x] Review/revise the PMIx `key` usage to be more memory footprint friendly, e.g., when used in a `pmix_info_t` structure (#367)
   - Revise PMIX_MAX_KEYLEN?
   - Convert `pmix_key_t` to a `char*` pointer?
  - No changes to these structures is being proposed.
- [x] Scrub the Standard doc to clarify who (application or PMIx library) retains responsibility for releasing memory in revised `pmix_proc_t` and `pmix_info_t` structures should they be modified from current definition - probably would need to be done on an API-by-API basis
[...]

PR #397, opened 2022-03-15
Add PMIX_QUERY_ABI_VERSION

jjhursey:
* Ref Issue #365
 * Add a `PMIX_QUERY_ABI_VERSION` attribute that can be passed to
   `PMIx_Query_info` to access the ABI version implementated by the
   associated PMIx library.

### DIFF
--- a/Chap_API_Query.tex
+++ b/Chap_API_Query.tex
@@ -356,13 +356,13 @@ Return an array of \refstruct{pmix_info_t}, each element itself containing a \re
 %
 \declareAttributeNEW{PMIX_QUERY_STABLE_ABI_VERSION}{"pmix.qry.stabiver"}{char *}{
 Query the PMIx Standard Stable \ac{ABI} version(s) supported by the \ac{PMIx} library. The version returned will be of the form \code{"MAJOR.MINOR"}. If multiple versions are supported then a comma-separated list of version numbers will be returned.
-%See Section~\ref{chap:intro:abi:versioning} for versioning advice.
+See Section~\ref{chap:intro:abi:versioning} for versioning advice.
 NO QUALIFIERS.
 }
 %
 \declareAttributeNEW{PMIX_QUERY_PROVISIONAL_ABI_VERSION}{"pmix.qry.prabiver"}{char *}{
 Query the PMIx Standard Provisional \ac{ABI} version(s) supported by the \ac{PMIx} library. The version returned will be of the form \code{"MAJOR.MINOR"}. If multiple versions are supported then a comma-separated list of version numbers will be returned.
-%See Section~\ref{chap:intro:abi:versioning} for versioning advice.
+See Section~\ref{chap:intro:abi:versioning} for versioning advice.
 NO QUALIFIERS.
 }
 
--- a/Chap_API_Query.tex
+++ b/Chap_API_Query.tex
@@ -144,6 +144,17 @@ Also can be used to query node-specific info such as the list of peers executing
 
 The returned \refarg{status} indicates if requested data was found or not. The returned \refarg{info} array will contain a \refattr{PMIX_QUERY_RESULTS} element for each query of the \refarg{queries} array. If qualifiers were included in the query, then the first element of each results array shall contain the \refattr{PMIX_QUERY_QUALIFIERS} key with a \refstruct{pmix_data_array_t} containing the qualifiers. The remaining \refstruct{pmix_info_t} shall contain the results of the query, one entry for each key that was found. Note that duplicate keys in the \refarg{queries} array shall result in duplicate responses within the constraints of the accompanying qualifiers. The caller is responsible for releasing the returned array.
 
+The \refapi{PMIx_Query_info} operation is expected to be called between initialization (e.g., \refapi{PMIx_Init}) and finalization (e.g., \refapi{PMIx_Finalize}) with the exeption of calling the blocking query \ac{API} exclusively with the following set of query attributes:
+\begin{compactitem}
+\item \refattr{PMIX_QUERY_ABI_VERSION}
+\end{compactitem}
+If this operation is called exclusively with one or more of those attributes then the result is computed locally to the \ac{PMIx} implementation and returns the associated value.
+
+\rationalestart
+Applications that wish to query the \ac{ABI} level supported by the \ac{PMIx} implementation wish to do so before calling any initialization routine.
+This allows such applications to query the \ac{ABI} level and take action based on that information which might include calling the initialization routine in a different manner and/or setting up the environment before the \ac{PMIx} implementation is initialized.
+\rationaleend
+
 \adviceimplstart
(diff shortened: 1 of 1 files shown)


jjhursey:
Left to do:
- [ ] Implement this in OpenPMIx for folks to experiment with.
- [ ] Post an example of how this attribute could be used. Maybe as a use case?

rhc54:
> Implement this in OpenPMIx for folks to experiment with.

The query_info functions currently rely on the proc having called `PMIx_Init` first. I'd suggest just having those functions start by doing a quick check for this attribute and responding immediately if found so you don't require the "init".

Chap_API_Sync_Access.tex
```diff
@@ -544,6 +544,17 @@ \subsection{\code{PMIx_Query_info}}
 
 The returned \refarg{status} indicates if requested data was found or not. The returned \refarg{info} array will contain a \refattr{PMIX_QUERY_RESULTS} element for each query of the \refarg{queries} array. If qualifiers were included in the query, then the first element of each results array shall contain the \refattr{PMIX_QUERY_QUALIFIERS} key with a \refstruct{pmix_data_array_t} containing the qualifiers. The remaining \refstruct{pmix_info_t} shall contain the results of the query, one entry for each key that was found. Note that duplicate keys in the \refarg{queries} array shall result in duplicate responses within the constraints of the accompanying qualifiers. The caller is responsible for releasing the returned array.
 
+The \refapi{PMIx_Query_info} and \refapi{PMIx_Query_info_nb} operations are expected to be called between initialization (e.g., \refapi{PMIx_Init}) and finalization (e.g., \refapi{PMIx_Finalize}) with the exeption of calling with exclusively with the following set of query attributes:
```
jjhursey, line 547:
```suggestion
The \refapi{PMIx_Query_info} operation is expected to be called between initialization (e.g., \refapi{PMIx_Init}) and finalization (e.g., \refapi{PMIx_Finalize}) with the exeption of calling the blocking query \ac{API} exclusively with the following set of query attributes:
```

jjhursey:
I filed the implementation of this in https://github.com/openpmix/openpmix/pull/2551 with some examples in the `examples/` directory.

jjhursey:
Refs:
 * OpenPMIx implementation: https://github.com/openpmix/openpmix/pull/2551
 * PMIx ABI Example: https://github.com/pmix/pmix-abi/pull/10

jjhursey:
[](https://github.com/)Please use emoji reactions ON THIS COMMENT to indicate your position on this proposal.

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

Chap_API_Sync_Access.tex
```diff
@@ -771,6 +782,10 @@ \subsection{Query attributes}
 \declareAttributeNEW{PMIX_SERVER_INFO_ARRAY}{"pmix.srv.arr"}{pmix_data_array_t}{
 Array of \refstruct{pmix_info_t} about a given server, starting with its \refattr{PMIX_NSPACE} and including at least one of the rendezvous-required pieces of information.
 }
+%
+\declareAttributeNEW{PMIX_QUERY_ABI_VERSION}{"pmix.qry.abiversion"}{char *}{
+Query the \ac{ABI} version supported by the \ac{PMIx} library. The version returned will be of the form \code{"MAJOR.MINOR"}. For example. \code{"4.2"} representing version v4.2 of the \ac{PMIx} \ac{ABI}. NO QUALIFIERS.
+}
```
jjhursey, line 788: Notes from April 14, 2022 meeting
 * Good discussion about how the "Major.Minor" will be updated and how an application might use this information.
 * What is the guidance on updating the "Major.Minor"?
   * If a new API is added, this does not break the ABI but does change the ABI. So increment the "Minor".
   * If an API is removed, this does break the ABI. So increment the "Major" and set "Minor" to 0.
   * If an API is modified, this does break the ABI. So increment the "Major" and set "Minor" to 0.
 * If an application builds against "`1.7`" of the ABI.
   * They require a major version of "`1`" and a minor version of "`x >= 7`" since they do not care about additions to the ABI only that the `1.7` ABI is still intact.
   * If the major version is not "`1`" then they need to consult the ABi to determine if the modifications/removals impact their application.
 * The complexity comes with Provisional items that are allowed to change form between minor revisions. This is in contrast to Stable items that have a longer Deprecation/Removal process.
   * If a Provisional API is changed then it changes the "Major" number in the ABI version causing broad disruption.
   * The suggestion was to separate the Stable and Provisional interfaces (as best as possible) and have two queries. `PMIX_QUERY_ABI_VERSION_STABLE` and `PMIX_QUERY_ABI_VERSION_PROVISIONAL`. This will help isolate applications focused on stable interfaces from rapid development of provisional items.
 * In some circumstances an implementation may continue to provide deprecated/removed interfaces. In this case they support multiple versions of the ABI. Consider two ABI versions `1.7` and `2.0` where the only difference is the removal of `PMIx_Foo()`
   * They support the old version `1.7` since it still includes `PMIx_Foo()`.
   * They also support the new version `2.0`. They just have an additional function `PMIx_Foo()` that is not defined in the `2.0` ABI.
   * As such it is important to allow an implementation to report more than one ABI level.
   * Suggestion: allow for a comma separated list to be returned.
   * If the difference between `1.7` and `2.0` is a modification of an existing API/structure then an implementation must determine which ABI they conform to. The implementation might choose to add a configure time option to choose the ABI level. Other runtime level options are available though more complex to implement.

jjhursey: FYI. I added this to the Governance language
 * https://github.com/pmix/governance/pull/34/files#r851349685
I'm going to suggest that we also add it to the PMIx Standard (in PR #375) then reference that section here

jjhursey: `pmix-abi` ticket to identify the Stable vs Provisional ABI:
 * https://github.com/pmix/pmix-abi/issues/11

jjhursey: I pushed a new commit to address this comment.

jjhursey:
I just pushed an additional commit to split the stable and provisional into two attributes.
The link to the versioning section is in a comment because it depends on PR #375 - we will address that at a later stage.

jjhursey:
PMIx ASC 2Q 2022
 * Passed the first vote: 8 yes / 0 no / 3 abstain
 * This is eligible for second vote in the 3Q 2022 ASC quarterly meeting

jjhursey:
Reference implementations:
 * OpenPMIx: https://github.com/openpmix/openpmix/pull/2617
   * Prior, earlier implementation: https://github.com/openpmix/openpmix/pull/2551
 * PMIx ABI Repo example: https://github.com/pmix/pmix-abi/pull/10

jjhursey:
PMIx ASC 3Q 2022
 * Passed the second vote: 8 yes / 0 no / 0 abstain
 * This will be included in the next major release.

jjhursey:
Rebased branch. The only new addition was the last commit (per discussion with the group) to add a cross-reference to the versioning language which was brought in as part of PR #375.

PR #398, opened 2022-04-04
IAWG: First pass at publish_lookup chapter

dsolt:
IAWG's first pass at changes to the Publish/Lookup chapter.  This is a WIP, but far enough along that people may be interested in seeing what we are working on.  Commits are not yet signed-off.

### DIFF
--- a/Chap_API_Publish.tex
+++ b/Chap_API_Publish.tex
@@ -199,16 +199,18 @@ Lookup information published by a process or host environment using \refapi{PMIx
 \cspecificstart
 \begin{codepar}
 pmix_status_t \\
-PMIx_Lookup_datastore(pmix_pdsdata_t data[], \\  
-\hspace*{12\sigspace}size_t ndata, \\
+PMIx_Lookup_datastore(pmix_key_t keys[], \\
+\hspace*{12\sigspace}pmix_pdsdata_t data[], \\  
+\hspace*{12\sigspace}size_t nkeys_and_data, \\
 \hspace*{12\sigspace}const pmix_info_t info[], \\
 \hspace*{12\sigspace}size_t ninfo);
 \end{codepar}
 \cspecificend
 
 \begin{arglist}
-\arginout{data}{Array of \refstruct{pmix_pdsdata_t} structures indicating the keys to lookup and providing storage for the results. (array of \refstruct{pmix_pdsdata_t})}
-\argin{ndata}{Number of elements in the \refarg{data} array (integer)}
+\argin{keys}{array of keys (array of strings)}
+\arginout{data}{Array of \refstruct{pmix_pdsdata_t} structures providing storage for the results. (array of \refstruct{pmix_pdsdata_t})}
+\argin{nkeys_and_data}{Number of elements in the \refarg{keys} and \refarg{data} arrays (integer)}
 \argin{info}{Array of info structures (array of handles)}
 \argin{ninfo}{Number of elements in the \refarg{info} array (integer)}
 \end{arglist}
@@ -248,11 +250,14 @@ A lookup operation is always performed on a range which can be specified using t
 The lookup operation will be constrained to data published to the specified range.
 Data is returned per the retrieval rules of Section \ref{chap:pub:retrules}.
 
-The \argref{data} parameter consists of an array of \refstruct{pmix_pdsdata_t} structures with the keys specifying the requested information.
-Data will be returned for each \code{key} field in the associated \code{value} and \code{publish_id} fields of this structure.
-The length of the arrays \code{publish_id} and \code{value} will be the number of published values matching the requested
-key and may be 0.  The \code{value} \refstruct{pmix_data_array_t} will have a \code{type} of \refstruct{pmix_value_t} 
-and the \code{publish_id} \refstruct{pmix_data_array_t} will have a \code{type} of \refstruct{pmix_publish_id_t}.
+The \argref{key} parameter consists of an array of \refstruct{pmix_key_t} structures specifying the requested information.
+The \argref{data} parameter provides space for the results.  The operation will fill in each 
+\argref{data} structure's key field with the corresponding key in the \argref{key} array. 
+The \code{value} and \code{publish_id} fields will be filled in with the results of the lookup operation.
+The length of the \code{publish_id} and \code{value} data arrays will be the number of published values matching the requested
+key and may be 0.  The \code{value} field of \refstruct{pmix_data_array_t} will contain \refstruct{pmix_value_t} elements and will have its \code{type} field set to \refconst{PMIX_VALUE}.
+The \code{publish_id} field of \refstruct{pmix_data_array_t} will contain \refstruct{pmix_publish_id_t} elements 
+and will have its \code{type} field set to \refconst{PMIX_PUBLISH_ID}.
 
 \adviceuserstart
 Although this is a blocking function, it will not wait by default for the requested data to be published.
@@ -278,16 +283,18 @@ Nonblocking version of \refapi{PMIx_Lookup_datastore}.
 \begin{codepar}
 pmix_status_t \\
 PMIx_Lookup_datastore_nb(pmix_key_t keys[], \\
+\hspace*{15\sigspace}size_t nkeys, \\
 \hspace*{15\sigspace}const pmix_info_t info[], size_t ninfo, \\
 \hspace*{15\sigspace}pmix_lookup_datastore_cbfunc_t cbfunc, void *cbdata);
 \end{codepar}
 \cspecificend
 
 \begin{arglist}
-\argin{keys}{\code{NULL}-terminated array of keys (array of strings)}
+\argin{keys}{array of keys (array of strings)}
+\argin{nkeys}{Number of elements in the \refarg{keys} array (integer)}
 \argin{info}{Array of info structures (array of handles)}
 \argin{ninfo}{Number of elements in the \refarg{info} array (integer)}
-\argin{cbfunc}{Callback function (handle)}
+\argin{cbfunc}{Callback function \refapi{pmix_lookup_datastore_cbfunc_t} (function reference)}
 \argin{cbdata}{Callback data to be provided to the callback function (pointer)}
 \end{arglist}
 
@@ -327,7 +334,8 @@ The following attributes are optional for host environments that support this op
 %%%%
 \descr
 
-Non-blocking form of the \refapi{PMIx_Lookup_datastore} function.
+Non-blocking form of the \refapi{PMIx_Lookup_datastore} function.  The \refapi{pmix_lookup_datastore_cbfunc_t} 
+will be called on successful completion and will provide available values for the requested keys.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \section{\code{PMIx_Unpublish_datastore}}
@@ -896,7 +904,7 @@ PMIx_Lookup_nb(char **keys, \\
 \argin{keys}{\code{NULL}-terminated array of keys (array of strings)}
 \argin{info}{Array of info structures (array of handles)}
 \argin{ninfo}{Number of elements in the \refarg{info} array (integer)}
-\argin{cbfunc}{Callback function (handle)}
+\argin{cbfunc}{Callback function \refapi{pmix_lookup_cbfunc_t} (function reference)}
 \argin{cbdata}{Callback data to be provided to the callback function (pointer)}
 \end{arglist}
 


Chap_API_Publish.tex
```diff
@@ -557,29 +557,32 @@ \section{Retrieval rules for published data}
...
-three values.
+A lookup of "tree" by P1 on the range \refconst{PMIX_RANGE_GLOBAL} does permit possible matches of keys published by P0.  P0 has published "tree" three times,
+but P1 is outside of the P0's node and namespace ranges.  Therefore, only the value of "tree"
+published to the session level is retrievable by P1.  Notice that the lookup range
+does not need to match the publishing range for a match to occur.  The requirement
+is that the requester is in the range specified by the publisher and the publisher is in the range specified by the requester.
+
+The figure shows a similar scenario for P2.  However, unlike P1, P2 is in the same namespace as P0 which results in two possible published values of
+"tree" that could be returned by a call to \refapi{PMIx_Lookup} by P2 on the range \refconst{PMIX_RANGE_SESSION} or \refconst{PMIX_RANGE_NAMESPACE}.
+An implementation is required to return one of the possible matches in such cases where
```
rhc54, line 584: Yeah, I think we have to preclude this possibility - it should be an error.

Chap_API_Publish.tex
```diff
@@ -557,29 +557,32 @@ \section{Retrieval rules for published data}
...
-    \item The lookup key matches the previously published key
-    \item The publishing process of the previously published key falls within the lookup range specified by the requester.
+    \item The lookup key matches the published key.
+    \item The publisher of the published key falls within the lookup range specified by the requester.
     \item The the requester falls within the range specified by the publisher.
     \item If the publisher specified access permissions, the effective \ac{UID} and \ac{GID} of the requester must meet those requirements.
 \end{enumerate}
 
-Fig.~\ref{fig:publish_lookup} shows process P0 executing on host X in namespace 1 in session A publishing the key "tree" on three ranges.  Publishing to different ranges is allowed.
+Fig.~\ref{fig:publish_lookup} shows process P0 executing on Host X in Namespace 1 in Session A publishing the key "tree" on three ranges.  Publishing to different ranges is allowed.
```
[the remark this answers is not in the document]
dsolt: I agree.  I was just trying to capture what I thought the document says currently, but I agree it is very complex to understand and to implement and it still has non-determinant cases (results that may vary based on implementation).   I'm all for making something that is as simple as possible both to understand and implement, even if that means it is not very rich in features.  Like you said, this is just meant to be a simple rendezvous mechanism.   I need to think more about what options we have and what we could do without breaking backward compatibility.  This is one of those cases where I don't think anyone is invested in the way things "work" now, but not sure how much liberty we should take on changing behavior.  

On another note, I would love to see the Publish API take 2 info arrays... one for the things to publish and one for modifiers to the call.  As it is now if I call Publish with an array of infos like:  

PMIX_RANGE=PMIX_RANGE_NAMESPACE
tree=oak
PMIX_RANGE=PMIX_RANGE_GLOBAL
tree=elm
PMIX_TIMEOUT=10

then we have several problems:  If the implementation doesn't handle or recognize PMIX_TIMEOUT then it will actually publish PMIX_TIMEOUT as a key since it doesn't recognize it as a directive.  Also, will it publish tree=oak at the namespace range and tree=elm at the Global range or will it publish only one of the values of tree at some range (either namespace or global)?   Things get really tricky when the things to publish are combined with the directives.

But deprecating existing API's for new Publish API's seems like a lot of work for something that very few products are using and probably have figured out a way to not get tripped up by these things already.  I don't want the pmix-standard to be a burden to implementations or progress, but I would like it to describe how things actually work so someone can develop new clients without getting lost or confused.   I'll see if I can think up some way to simplify what is supported without breaking backward compatibility.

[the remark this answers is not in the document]
rhc54: So don't deprecate the API, but add an attribute for specifying modifiers to the call. It would take a `pmix_data_array_t` containing an array of `pmix_info_t` that has the modifiers in them. I don't think that would be much of a problem for backwards compatibility.

If you prefer, you could add an attribute for the things to publish, leaving all else in the call to be modifiers. Either way should be fine.

I wouldn't worry too much about changes in behavior at this point. OMPI is the primary user of it (for connect/accept), and I've already outlined a way for them to use `PMIx_Group` instead - which is a far better and more performant solution for that use-case.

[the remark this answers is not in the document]
dsolt: I like that idea.   We could have both approaches for a while, then move towards deprecating the old way after a while.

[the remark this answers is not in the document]
rhc54: Thinking some more, we should probably add the attribute for the things to publish (as opposed to the modifiers) so we remain consistent with the other APIs - whose info arrays all contain modifiers.

Chap_API_Publish.tex
```diff
@@ -557,29 +557,32 @@ \section{Retrieval rules for published data}
...
-The key being looked up will match with a previously published key only if all of the following conditions are met:
+All publish and lookup operations operate on a range. If not specified, the range defaults to \refconst{PMIX_RANGE_SESSION}.
+The key being looked up will match with a published key only if all of the following conditions are met:
 
 \begin{enumerate}
-    \item The lookup key matches the previously published key
-    \item The publishing process of the previously published key falls within the lookup range specified by the requester.
+    \item The lookup key matches the published key.
+    \item The publisher of the published key falls within the lookup range specified by the requester.
     \item The the requester falls within the range specified by the publisher.
```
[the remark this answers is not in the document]
dsolt: got it.  Will show up in next push.

Chap_API_Publish.tex
```diff
@@ -73,7 +72,8 @@ \section{\code{PMIx_Publish}}
 The blocking form of this call will block until it has obtained confirmation from the datastore that the data is available for lookup. The \refarg{info} array can be released upon return from the blocking function call.
 
 Publishing duplicate keys is permitted provided they are published to different
-ranges. Duplicate keys being published on the same data range shall return the
+ranges. Custom ranges are consider different if they have different members.
```
naughtont3, line 75:
```suggestion
ranges. Custom ranges are considered different if they have different members.
```

Chap_API_Struct.tex
```diff
@@ -2253,6 +2253,40 @@ \subsection{Release Callback Function}
...
+%%%%
+\descr
+
+A callback function for calls to \refapi{PMIx_Lookup_nb}.
+The function will be called upon completion of the \refapi{PMIx_Lookup_nb} \ac{API} with the \refarg{status} indicating the success or failure of the request.
+Any retrieved data will be returned in an array of \refstruct{pmix_pdata_t} structs.
+The namespace and rank of the process that provided each data element is also returned.
+
+Note that the \refstruct{pmix_pdata_t} structures will be released upon return from the callback
+function, so the receiver must copy/protect the data prior to returning if it needs to be retain ed.
```
naughtont3, line 2288:
```suggestion
function, so the receiver must copy/protect the data prior to returning if it needs to be retained.
```

dsolt:

    we know your current thoughts

    want to be an advocate for it
    other technologies instead

jjhursey:
@dsolt Are all of the comments in this PR resolved? (If so then can you mark them as resolved?)

Chap_API_Publish.tex
```diff
@@ -195,25 +230,33 @@ \subsection{Range of Published Data}
 Undefined range.
 %
 \declareconstitemvalue{PMIX_RANGE_RM}{1}
-Data is intended for the host environment, or lookup is restricted to data published by the host environment.
+Published data and generated events are restricted to processes executing under the same instance of the host environment as the publisher or event creator.  Lookup of data is restricted to data published by processes running under the same instance of the host environment as the requester.
 %
 \declareconstitemvalue{PMIX_RANGE_LOCAL}{2}
```
abouteiller, line 235: Reading this text I stumbled again on the question of wether ranges are "nested" or not. The new text in `PMIx_Lookup` below makes it very clear that it is not, but some forewarning here (or a ref to the lookup rules  `\ref{chap:pub:retrules}.`) could help clarify.

Chap_API_Publish.tex
```diff
@@ -542,59 +591,28 @@ \subsubsection{Lookup data structure support macros}
...
-    \item If the key of the stored information does not match the specified key, then the search will continue.
-
-    \item If the requester's identifier does not fall within the range specified by the publisher, then the search will continue.
-
-    \item If the publisher specified access permissions, the effective \ac{UID} and \ac{GID} of the requester shall be checked against those permissions, with the datastore rejecting the match if the requester fails to meet the requirements.
-
-    \item If all of the above checks pass, then the value is added to the information that is to be returned.
+    \item The lookup key matches the published key.
+    \item The type of range specified by the publisher is the same as the type of range specified by the requester.
+    \item If a custom range is specified by the publisher and the requester, the members described in both cases must be identical.  The publisher is not required to be a member of a custom range.
```
abouteiller, line 605: This seems to contradict the rule from https://github.com/pmix/pmix-standard/pull/398#discussion_r940372741

Chap_API_Publish.tex
```diff
@@ -74,9 +76,34 @@ \section{\code{PMIx_Publish}}
...
 
 Publishing duplicate keys is permitted provided they are published to different
-ranges. Duplicate keys being published on the same data range shall return the
+ranges. Custom ranges are consider different if they have different members.
+Duplicate keys being published on the same data range shall return the
 \refconst{PMIX_ERR_DUPLICATE_KEY} error.
-
+publishing to a \refconst{PMIX_RANGE_CUSTOM}
+range which does not include the publisher will prevent
+any processes from using \refapi{PMIx_Lookup} to access the published data.
```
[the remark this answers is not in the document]
rhc54: I'm actually not sure that statement is correct - the range determines who can access it. I'm not sure why the publisher would need to be a member of that group. The publisher always "owns" the data and can "unpublish" it.

jjhursey:
For the meeting today, attached is a PDF rendering of the current state of this PR.
 * [pmix-standard-3q-2022-pr-398.pdf](https://github.com/pmix/pmix-standard/files/9291477/pmix-standard-3q-2022-pr-398.pdf)

Chap_API_Publish.tex
```diff
@@ -74,9 +76,34 @@ \section{\code{PMIx_Publish}}
 The blocking form of this call will block until it has obtained confirmation from the datastore that the data is available for lookup. The \refarg{info} array can be released upon return from the blocking function call.
 
 Publishing duplicate keys is permitted provided they are published to different
-ranges. Duplicate keys being published on the same data range shall return the
+ranges. Custom ranges are consider different if they have different members.
+Duplicate keys being published on the same data range shall return the
 \refconst{PMIX_ERR_DUPLICATE_KEY} error.
-
+publishing to a \refconst{PMIX_RANGE_CUSTOM}
```
jjhursey, line 82:
```suggestion
Publishing to a \refconst{PMIX_RANGE_CUSTOM}
```

rhc54:
I worked thru the chapter and made some recommended changes
Ralph

Chap_API_Publish.tex
```diff
@@ -74,9 +76,34 @@ \section{\code{PMIx_Publish}}
...
+In some cases, implementations may be incapable of distinguishing which
+info keys in the \refarg{info} array are for publishing and which info keys are
+directives.  To make it clear, it is recommended that the keys to be published
+are designated by passing them as a \refstruct{pmix_data_array_t} using the
+\refattr{PMIX_DATA_TO_PUBLISH} directive.
+If the \refarg{info} array contains a \refattr{PMIX_DATA_TO_PUBLISH} info,
+all other elements of the info array will be treated as directives.
+If the info array does not include a \refattr{PMIX_DATA_TO_PUBLISH} info,
+the implementation should
+distinguishing between info array elements that specify keys and directives as follows:
```
jjhursey, line 95:
```suggestion
the implementation should
distinguish between info array elements that specify keys and directives as follows:
```

Chap_API_Publish.tex
```diff
@@ -542,59 +591,28 @@ \subsubsection{Lookup data structure support macros}
...
-
-    \item If the requester's identifier does not fall within the range specified by the publisher, then the search will continue.
-
-    \item If the publisher specified access permissions, the effective \ac{UID} and \ac{GID} of the requester shall be checked against those permissions, with the datastore rejecting the match if the requester fails to meet the requirements.
-
-    \item If all of the above checks pass, then the value is added to the information that is to be returned.
+    \item The lookup key matches the published key.
+    \item The type of range specified by the publisher is the same as the type of range specified by the requester.
+    \item If a custom range is specified by the publisher and the requester, the members described in both cases must be identical.  The publisher is not required to be a member of a custom range.
+    \item The requestor must be a member of the publisher's range.
```
jjhursey, line 606:
```suggestion
    \item The requestor must be a member of the range specified by the publisher.
```

Chap_API_Publish.tex
```diff
@@ -542,59 +591,28 @@ \subsubsection{Lookup data structure support macros}
...
-
-    \item If the key of the stored information does not match the specified key, then the search will continue.
-
-    \item If the requester's identifier does not fall within the range specified by the publisher, then the search will continue.
-
-    \item If the publisher specified access permissions, the effective \ac{UID} and \ac{GID} of the requester shall be checked against those permissions, with the datastore rejecting the match if the requester fails to meet the requirements.
-
-    \item If all of the above checks pass, then the value is added to the information that is to be returned.
+    \item The lookup key matches the published key.
+    \item The type of range specified by the publisher is the same as the type of range specified by the requester.
```
jjhursey, line 604: Is this required (e.g., requestor specifies 'node' range, and publisher specifies 'global' range)? Or is it enough that the requestor be part of the publisher's range?

Chap_API_Publish.tex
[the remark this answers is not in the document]
dsolt: Good catch.  At one point in our revision process we had determined that the publisher and process calling Lookup both had to be part of the group, but we decided to remove that restriction and allow the publisher to not be part of the group of processes, so this should have been removed when we made that switch.

Chap_API_Publish.tex
dsolt: Yes, this is correct.  The other location is wrong.

Chap_API_Publish.tex
dsolt: now a nop

Chap_API_Publish.tex
dsolt: done

Chap_API_Publish.tex
dsolt: The new model is that the ranges must match... if you publish to 'global', then the requestor must specifically lookup on 'global' range.

Chap_API_Publish.tex
```diff
@@ -74,9 +76,31 @@ \section{\code{PMIx_Publish}}
...
+distinguish between info array elements that specify keys and directives as follows:
+All standardized directives to the publish call,
+including optional attributes the implementation does not support,
+should be treated as
+directives.  Non-supported directives
+may be ignored as outlined in Section \ref{intro:portability:attributes},
+but should not be treated as data to
+publish.  The implementation may treat any custom (non-standardized) directives it
+supports as directives.  All other \refarg{info} array elements
+should it be assumed to be data to be published.
```
jjhursey, line 101:
```suggestion
should be assumed to be data to be published.
```

Chap_API_Publish.tex
```diff
@@ -74,9 +76,31 @@ \section{\code{PMIx_Publish}}
 The blocking form of this call will block until it has obtained confirmation from the datastore that the data is available for lookup. The \refarg{info} array can be released upon return from the blocking function call.
 
 Publishing duplicate keys is permitted provided they are published to different
-ranges. Duplicate keys being published on the same data range shall return the
+ranges. Custom ranges are consider different if they have different members.
+Duplicate keys being published on the same data range shall return the
 \refconst{PMIX_ERR_DUPLICATE_KEY} error.
 
+In some cases, implementations may be incapable of distinguishing which
```
jjhursey, line 83: Note from 3Q 2022 ASC (Aug. 11, 2022):
 - Let's require `PMIX_DATA_TO_PUBLISH` as the only way to publish data.
   - This has some backward compatibility issues.
     - Open MPI and (maybe) MPICH would be impacted
   - We should add an error if `PMIX_DATA_TO_PUBLISH` is not provided in the publish call.
   - In v5, mark not using `PMIX_DATA_TO_PUBLISH` as deprecated.
 - Should we just create a new API with an additional required parameter for the data_to_publish (what to name it? `PMIx_Publish_data`?)
   - Then we just deprecate the old API instead of adding a new attribute to work around it.
```c
pmix_status_t PMIx_Publish_NEW(const pmix_info_t info_to_publish[], size_t ninfos,
                               const pmix_info_t info_directives[], size_t ndirs);
```

Chap_API_Publish.tex
```diff
@@ -195,25 +227,33 @@ \subsection{Range of Published Data}
...
 \declareconstitemvalue{PMIX_RANGE_PROC_LOCAL}{7}
-Data is only available to this process, or lookup is restricted to data published by this process.
+Published data and generated events are available only to calling process.
+Lookup of data is restricted to data published by the calling process.
+%
+\declareconstitemvalue{PMIX_RANGE_CUSTOM}{6}
+Published data and generated events are restricted to processes
+described in the \refstruct{pmix_info_t} associated with this call.
+Lookup of data is restricted to data published by the processes described in
+in the \refstruct{pmix_info_t}.
```
jjhursey, line 256: Note from 3Q 2022 (Aug. 11, 2022)
 - What does it mean for two `PMIX_RANGE_CUSTOM`s to match? Is it a strict ordering or is a subset fine?
 - Is the publisher required to be part of the range?
   - Publisher does need to be a member of the range?
   - "If I publish that data I own that data and can look it up so that I can unpublish"
     - Current description would prevent the lookup but not the unpublish.
     - Owner should always be able to look it up even if not in the range. Maybe need a clause here? What about overlapping keys?
 - Publish means "put some data for a set of consumers to lookup". The consumers might not know who published the data. For a custom range, those looking up might not know the full group. Alternatively, the lookup could specify a list including the publisher.

Chap_API_Publish.tex
```diff
@@ -4,21 +4,20 @@
...
+information that either is associated with a specific process (i.e., the
+retrieving process knows the identifier of the process that posted it) or
+requires a synchronization operation prior to retrieval (e.g., the case of
 globally unique non-reserved keys). However, another requirement exists for an
 asynchronous exchange of data where neither the posting nor the retrieving
-process is known in advance. For example, two separate namespaces may need to
-rendezvous with each other without knowing in advance the identity of the other
-namespace or when that namespace might become active.
-
+process is known in advance, for example, two namespaces that do not share a child-parent relationship.
```
jjhursey, line 16:
```suggestion
process is known in advance (e.g., two namespaces that do not share a child-parent relationship).
```

Chap_API_Publish.tex
```diff
@@ -195,25 +227,33 @@ \subsection{Range of Published Data}
 Undefined range.
 %
 \declareconstitemvalue{PMIX_RANGE_RM}{1}
-Data is intended for the host environment, or lookup is restricted to data published by the host environment.
+Published data and generated events are restricted to processes executing under the same instance of the host environment as the publisher or event creator.  Lookup of data is restricted to data published by processes running under the same instance of the host environment as the requester.
```
jjhursey, line 230: Note from 3Q 2022 (Aug. 11, 2022)
 * Old text was intended. That the published data is intended solely for the host environment not for application/tool processes.
   * Suggest leave the old text
 * `PMIX_RANGE_GLOBAL` : Anyone under that RM instance (basically what the new text is saying)

jjhursey:
3Q 2022:
 * Reading turned up a few items that need further discussion.
 * Will follow up in the IAWG to file tune the last few items.
 * Intention to bring this back for presentation during the next quarterly meeting.

Chap_API_Publish.tex
jjhursey: Per IAWG on Aug. 15, 2022:
 * Add new `PMIx_Publish2` and `pmix_server_publish2_fn_t` with two sets of `pmix_info_t`s
 * Deprecate the `PMIx_Publish` API and corresponding `pmix_server_publish_fn_t` API.
 * Do not add the `PMIX_DATA_TO_PUBLISH` attribute. Instead, prefer the new API.
 * In the Revision appendix note, `PMIx_Publish` was deprecated and replaced by `PMIx_Publish2` (same with server API)
 * No longer an issue on the ordering of info data and directives (this approach avoids this whole issue)

Chap_API_Publish.tex
jjhursey: Per IAWG on Aug. 15, 2022:
 * Does the publisher have to be part of the range it publishes to?
   * The publisher does not need to be a member of the range.
   * The publisher is the owner of the data, so it is the only one able to unpublish the data, but must specify the range where that data was published.
   * The publisher is the owner of the data but cannot lookup the data if it is not part of the range in which it was pushed.
 * If I'm looking up from a `PMIX_RANGE_CUSTOM` does it have to match what the publisher provided?
   * `PMIX_RANGE_NAMESPECE` (example)
     * **Publish**: means that the data is to be available to a set of lookup'ers in my namespace.
     * **Lookup**: means that I want data provided by a publisher in my namespace.
   * `PMIX_RANGE_CUSTOM`
     * **Publish**: 'custom' means I want the data to be available to this set of lookup'ers. Might not include the publisher.
     * **Lookup**: 'custom' means I want data provided by one of these publishers. Does not need to know all of the processes that have access to the data - only who is providing the data.
   * Generally `PMIX_RANGE_` means:
     * **Publish**: Define who can look up the data
     * **Lookup**: Define who published the data
 * Does the range have to match on the publish and lookup?
   * I'm publishing to range X, and I'm looking up on range X.
   * `PMIX_RANGE_CUSTOM` is the only one that takes an argument list of processes. It's the only one where the publisher and lookup'er might be using the same value associated with the `PMIX_RANGE_CUSTOM` key.
 * What if the publisher publishes the same key (different values) to different custom ranges that both include the client looking up the key? Or two different publishers both publish the same key (different values) that include the lookup client in different `PMIX_RANGE_CUSTOM`s.
   * This is not allowed.
   * The Server should reject the lookup operation as conflicting entries. The client should then change their custom range to include one
 * What If I want to lookup a value published by a process outside of my namespace?
   * Publisher will use `PMIX_RANGE_SESSION` or `PMIX_RANGE_CUSTOM` (if they know the lookup'er)
   * Lookup'er will user `PMIX_RANGE_CUSTOM` with the publisher's name
 * **Future work**: Maybe extend `PMIX_RANGE_NAMESPACE` with an argument of which namespace. Currently, it is the same as the caller.

Chap_API_Publish.tex
jjhursey: Per IAWG on Aug. 15, 2022:
 * We should make clean that though the ranges are overlapping in definition, they are not nested. So ranges must match.
 * We should add the forward reference in the top paragraph of this section to the lookup section where this is more clearly noted.

dsolt:
Here is the latest draft with the changes from the first reading (i.e. introduce a PMIx_Publish2, get rid of PMIX_DATA_TO_PUBLISH, and specify that the range of lookup is used to determine eligible publishers).

Will update the PR soon after WG has had the chance to review.

[pmix-standard.pdf](https://github.com/pmix/pmix-standard/files/9629548/pmix-standard.pdf)

jjhursey:
PMIx ASC 4Q 2022 Meeting presented a straw poll regarding:
 > Advice to the working group on how best to proceed with the Lookup semantics related to the range specified by the lookup and the scenario where there are duplicate keys published in separate ranges.

Three options were presented:
 * **Option 1**: Change the lookup range to specify where the data was published, not by whom it was published.
 * **Option 2**: If no range is specified in the lookup and multiple matches for a key exist in different ranges, then return a specific status code (e.g., PMIX_MULTIPLE_MATCH) and the full set of matching keys with associated range information.
 * **Option 3**: Do not allow publishing on ranges other than the publisher's namespace or global. Lookup would specify one or the other range.
 * Other

Sorted by choice preference
Option   | First Choice | Second Choice | Third Choice |    No | Abstain |
-------- | ------------ | ------------- | ------------ | ----- | ------- | 
Option 2 |            **6** |             3 |            2 |     1 |       0 |
Option 1 |            1 |             **8** |            1 |     1 |       1 |
Option 3 |            3 |             0 |            **4** |     4 |       1 |
Other    |            3 |             0 |            0 |     1 |       8 |

Other comments:
 * "I'd like to present a modified form of Option 2 that might provide an acceptable landing spot."
 * "Seemed like Option-3 required users to encode into key to make it functional, so that is why I abstained on that item, b/c I was not entirely sure that was satisfactory option."
 * "Use the approach given in the current PR which requires the lookup process to know the namespace of the publisher."
 * "If options 1 and 2 are compatible, then I think the best case is to do both."

jjhursey:
PMIx ASC 4Q 2022 day 2 meeting discussed this further and outlined a path forward. See the notes below
 * https://github.com/pmix/pmix-standard/wiki/ASC-Q4-2022-Meeting#day-2-oct-27-2022

jjhursey:
Notes from IAWG meeting Oct. 31, 2022
  - Introduce new APIs for Publish/Lookup/Unpublish (suffix with `_data` / `_ds` / `_datastore`)
    - Leave old API as they are with all ambiguities
  - The same key can be published by different or the same publisher into the datastore
  - **Mechanism for lookup**
    - Filter by access rules
    - Filter by source rules
    - Filter to find key
    - More than one value may be returned as an array of values (with qualifiers)
  - **Data store unique tuple**
    - `<key, value, publisher, access permissions, timestamp, handler id>`
  - **Publish**
    - Put a key/value pair into the datastore
    - Access permissions: Can specify the range of recipients that can access it (range)
      - Maybe we have a default?
    - Publisher can publish the same key multiple times - accumulates in the data store (duplicates allowed)
      - If a publisher publishes the same thing 10 times, then the lookup sees all 10 values
      - FUTURE WORK: Could add a 'REPLACE' attribute to replace given the handle?
    - Handed back a handler/id to reference when unpublishing
    - No limitations beyond this
  - **Lookup**
    - Ask for a 'key'
    - Filtered by access rules defined by the publisher
    - Source rules: Can specify the range of publishers from which to get the data
    - Since multiple matches can happen
      - If multiple matches for a 'key,' then you get multiple values back with qualifiers
      - Always get back an array - even if it is an array of one item
      - Qualifiers: publisher, timestamp, access permissions, others?
  - **Unpublish**
    - Only the original publisher can unpublish
    - Pass the handle from publishing to reference the data that was published
      - FUTURE WORK: Maybe add a qualifier to unpublish anything from me with this key (may match more than one)
    - Add 'access permissions'/range to modify the access permissions?

dsolt:
I added a new commit for:

* lookup blocking and non-blocking are so different. Should we separate keys from return values in the blocking call so it looks more like the non-blocking call.

jjhursey:
PMIx ASC 1Q 2023 Pleanary Notes (see [Notes Day 1](https://github.com/pmix/pmix-standard/wiki/ASC-Q1-2023-Meeting#day-1-jan-24-2023) for more details)
 * Noted some use cases
   * **Rendevous with failed Publisher processes**
      - Publisher A publishes 'find_me_key'
      - Publisher A dies
      - Publisher B takes over Publisher A's role
      - Publisher B unpublishes 'find_me_key'
      - Publisher B publishes 'find_me_key'
    * **Published key for the first consumer**
      - Publisher A publishes 'for_you'
      - Consumer Z looks up 'for_you'
      - Consumer Z unpublishes 'for_you'
        - Make it automatic on first lookup - so only 1 gets it
      - Publisher and Consumer must have the same uid/gid (owner)
    * **Lookup should have a sense of consistency**
      - Publisher A publishes 'key'
      - Consumer Z lookup 'key'
      - Publisher B publishes 'key'
      - Consumer Y lookup 'key'
      - Consumers must see at least Publisher A's key
 * The IAWG will need to have a reference implementation before voting.
   * Once semantics are ironed out then someone will need to work on the implementation
 * Is the old publish compatible with the new publish_datastore?
   * No, they would be separate storages and not share information. This is for the simplicity of the RM implementation and the semantics from the client side.
 * On lookup, how do you know which scope the value was published in?
   * Add a scope attribute on the lookup to define the scope. There will be a default.
   * You only get values exactly at that specified level.
   * Scopes are not nested. If not published at the 'session' level you won't see it even if it was published at the 'namespace' level.
 * Next Step
   * Finish the text for the presentation in the next quarterly
   * Discussion of implementation

PR #407, opened 2022-05-24
Issue 175 make query chap -rebased on master

dsolt:
Rebased version of pr#391. This is the one we actually want to merge, but we need to make sure it is applying the same changes as 391.

PR #412, opened 2022-05-26
Revision section for v5

jjhursey:


jjhursey:
* In PR #407 we removed `PMIX_QUERY_PARTIAL_SUCCESS`
https://github.com/pmix/pmix-standard/blob/04cc036e40f3185bed1cc25ca14c88d28ee0605c/Chap_API_Query.tex#L256-L268

I wanted to capture that removal in the Revision History and noticed that we don't have a v5 section. The RMs will need to fill this out based on what has been merged to `master` and not to `v4`. This should get you all started with a framework.

jjhursey:
@raffenet @dsolt Can you give this a review so we can merge?

PR #430, opened 2022-12-12
Add const to string parameters not meant to be modified

### LINKED ISSUE #426: Add const qualifier to all strings not meant to be modified
### Overview

In the signatures of some functions `char *` is used as type for strings clearly not meant to be modified by the function. In such cases in general the `const` qualifier should be added.

### Motivation

For example using the functions `PMIx_server_define_process_set()` and `PMIx_server_delete_process_set()` with a `const char*` string leads to a GCC compiler warning [-Wdiscarded-qualifiers]. This is avoidable and should be done, since this function clearly should not alter the string, most probably it will just copy it.

In other cases like that (e.g. `PMIx_Info_list_add()`) `const` is already in place, so perhaps it has just been missed.

### Discussion Items

To be honest, I don't see much room for discussion here. Would be interesting, why this not already happened, whether I am the first one struggling on this and if there is a good reason not to do this, that I cannot see.

A (hopefully) complete list of functions where I see the issue:
- `PMIx_Get_attribute_string(char *attributename)`
- `PMIx_Get_attribute_name(char *attributename)`
- `PMIx_Data_print(char **output, char *prefix, void *src, pmix_data_type_t type)`
- `PMIx_Register_attributes(char *function, pmix_regattr_t attrs[], size_t nattrs)`
- `PMIx_server_define_process_set(const pmix_proc_t members[], size_t nmembers, char *pset_name)`
- `PMIx_server_delete_process_set(char *pset_name)`

raffenet:
Without const, users may encounter `-Wdiscarded-qualifier` warnings at build time. While these changes modify API signatures, they should be regarded as ABI compatible. Fixes #426.

### DIFF
--- a/Chap_API_Data_Mgmt.tex
+++ b/Chap_API_Data_Mgmt.tex
@@ -332,13 +332,13 @@ Pretty-print a data value.
 
 \copySignature{PMIx_Data_print}{2.0}{
 pmix_status_t \\
-PMIx_Data_print(char **output, char *prefix, \\
+PMIx_Data_print(char **output, const char *prefix, \\
 \hspace*{16\sigspace}void *src, pmix_data_type_t type);
 }
 
 \begin{arglist}
 \argin{output}{The address of a pointer into which the address of the resulting output is to be stored. (\code{char**})}
-\argin{prefix}{String to be prepended to the resulting output (\code{char*})}
+\argin{prefix}{String to be prepended to the resulting output (\code{const char*})}
 \argin{src}{A pointer to the memory location of the data value to be printed (handle)}
 \argin{type}{The type of the data value to be printed --- must be one of the PMIx defined data types. (\refstruct{pmix_data_type_t})}
 \end{arglist}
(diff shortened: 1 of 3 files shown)


rhc54:
Don't forget the "pretty-print" functions - e.g., `PMIx_Error_string`

rhc54:
Never mind - you got the ones with pointers!

raffenet:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

jjhursey (approved):

jjhursey:
This was missed in the latest ASC meeting. If this is ready for inclusion, please notify the pmix-forum mailing list to get it on the 3Q 2023 agenda.

raffenet:
> This was missed in the latest ASC meeting. If this is ready for inclusion, please notify the pmix-forum mailing list to get it on the 3Q 2023 agenda.

We did read this at the Q1 meeting, but I can't seem to find record of a vote on it. Do we normally vote on errata at the same meeting they are read?

jjhursey:
Thanks for the note. it looks like we forgot to update the PR with the vote and merge.

1Q 2023 Voting results
 * 9 Yes / 0 No / 0 Abstain
 * Passed, and can be merged. Target v5.1

PR #433, opened 2023-01-24
Fix some print statements and update OpenPMIx header parsing

rhc54:
We are using Python3, but the print statements are a mix of
Python2 and 3 formats - so let's get everything into Python3.

The OpenPMIx header includes some functions that return
types (e.g., pmix_byte_object_t*) - the current parsing logic
ignores those entries, so let's update to include them.
This includes the following formats:
```text
const <type>* <fn>
<type>* <fn>
<type>** <fn>
<type> *<fn>
<type> **<fn>
```
This picked up a number of previously overlooked exceptions.

### DIFF
--- a/bin/check-openpmix.py
+++ b/bin/check-openpmix.py
@@ -12,9 +12,9 @@ def check_missing_pmix_standard(std_all_refs, openpmix_all_refs,
                                 openpmix_deprecated, verbose=False):
     """Check for OpenPMIx definitions missing from the PMIx Standard"""
 
-    print "-"*50
-    print "Checking: Defined in OpenPMIx, but not in the PMIx Standard"
-    print "-"*50
+    print("-"*50)
+    print ("Checking: Defined in OpenPMIx, but not in the PMIx Standard")
+    print ("-"*50)
 
     missing_refs = []
     for openpmix_ref in openpmix_all_refs:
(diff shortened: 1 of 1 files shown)


rhc54:
@jjhursey Sorry, I cannot get the `signed-off-by` checker to report its status. Tried to redeliver it multiple times.

bin/check-openpmix.py
```diff
@@ -81,11 +82,12 @@ def check_missing_openpmix(std_all_refs, openpmix_all_refs,
             p.wait()
             if p.returncode != 0 and p.returncode != 1:
                 print("Error: Failed to grep and confirm missing item \""+std_ref+"\". grep error code "+str(p.returncode)+")");
-                sys.exit(2)
+                continue
```
jjhursey, line 85: Don't we still want this to error out? This indicates that something is wrong with the script so the results are suspect.

bin/check-openpmix.py
```diff
@@ -114,7 +116,7 @@ def check_missing_openpmix(std_all_refs, openpmix_all_refs,
                         possible_error = True
 
                 if possible_error is True:
-                    sys.exit(1)
+                    continue
```
jjhursey, line 119: Don't we still want this to error out? This indicates that something is wrong with the script so the results are suspect.

jjhursey (changes requested):
Other than the two `continue` changes I think this is fine

bin/check-openpmix.py
rhc54: I wasn't entirely clear as to the reason for the error here. If it truly is something wrong with the script, then yes, we should error out.

bin/check-openpmix.py
rhc54: Ditto here - just wasn't clear to me that this was an error in the script.

rhc54:
Updated per comments

jjhursey (approved):

PR #443, opened 2023-02-27
Minor edits to Events chapter

dsolt:


Chap_API_Event.tex
```diff
@@ -43,7 +43,10 @@ \section{Notification and Management}
 %
 \end{itemize}
 
-Users can specify the callback order of a handler within its category at the time of registration. Ordering can be specified by providing the relevant event handler names, if the user specified an event handler name when registering the corresponding event. Thus, users can specify that a given handler be executed before or after another handler should both handlers appear in an event chain (the ordering is ignored if the other handler isn't included). Note that ordering does not imply immediate relationships. For example, multiple handlers registered to be serviced after event handler \textit{A} will all be executed after \textit{A}, but are not guaranteed to be executed in any particular order amongst themselves.
+Users can specify the callback order of a handler within its category at the time of registration. 
+Users can specify that a given handler be executed before or after another target handler should both handlers appear in an event chain (the ordering is ignored if the other handler isn't included). 
+The ordering is dictated by providing the event handler name of the target.  The name must have been assigned the target handler was registered.
```
abouteiller, line 48:
```suggestion
The ordering is dictated by providing the target event handler name.  The name must have been assigned when the target handler was registered.
```

Chap_API_Event.tex
```diff
@@ -375,7 +378,7 @@ \subsection{Notification Function}
 \adviceuserend
 
 \advicermstart
-On the server side, the notification function is used to inform the \ac{PMIx} server library's host of a detected event in the \ac{PMIx} server library. Events generated by \ac{PMIx} clients are communicated to the \ac{PMIx} server library, but will be relayed to the host via the \refapi{pmix_server_notify_event_fn_t} function pointer, if provided.
+On the server side, the notification function, \refapi{pmix_server_notify_event_fn_t} (See \ref{chap:server:func_pointers}), is used to inform the \ac{PMIx} server library's host of a detected event in the \ac{PMIx} server library. Events generated by \ac{PMIx} clients are communicated to the \ac{PMIx} server library, but will be relayed to the host via this \refapi{pmix_server_notify_event_fn_t} function pointer, if provided.
```
abouteiller, line 381:
```suggestion
On the server side, the notification function, \refapi{pmix_server_notify_event_fn_t} (See \ref{chap:server:func_pointers}), is used to inform the \ac{PMIx} server library's host of a detected event in the \ac{PMIx} server library. Events generated by \ac{PMIx} clients are communicated to the \ac{PMIx} server library, but will be relayed to the host via its \refapi{pmix_server_notify_event_fn_t} function pointer, if provided.
```

Chap_API_Event.tex
```diff
@@ -565,10 +570,16 @@ \subsubsection{Completion Callback Function Status Codes}
 %
 \declareconstitemvalue{PMIX_EVENT_ACTION_DEFERRED}{-333}
 Event handler: Action deferred.
+\end{constantdesc}
 %
-\declareconstitemvalue{PMIX_EVENT_ACTION_COMPLETE}{-334}
-Event handler: Action complete.
+
+The following status code may be returned by a handler to stop the handler chain.  It will not appear in the \refarg{results} for any handler invocation since it is the last handler executed in the chain.
```
abouteiller, line 576:
```suggestion
The following status code may be returned by a handler to stop the handler chain.  It will not appear in the \refarg{results} for any handler invocation since the handler that produces it is the last handler executed in the chain.
```

abouteiller (changes requested):
Looking good overall, I put some changes

Chap_API_Event.tex
dsolt: Ty and commited.

Chap_API_Event.tex
dsolt: Ty and commited

Chap_API_Event.tex
dsolt: The wording on the first part was intentional, though your suggestion may still be valid.  The reason we said it the way we did "event handler name of the target" was because event handlers may not have names, so we wanted it to be especially clear that you are providing the name of the target (which you would need to define) as opposed to "target event handler name" which would more easily make the reader believe that such a name exists.  Maybe your approach is clear enough though and certainly reads better, so we will discuss next week.

dsolt:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

Chap_API_Event.tex
```diff
@@ -43,7 +43,10 @@ \section{Notification and Management}
 %
 \end{itemize}
 
-Users can specify the callback order of a handler within its category at the time of registration. Ordering can be specified by providing the relevant event handler names, if the user specified an event handler name when registering the corresponding event. Thus, users can specify that a given handler be executed before or after another handler should both handlers appear in an event chain (the ordering is ignored if the other handler isn't included). Note that ordering does not imply immediate relationships. For example, multiple handlers registered to be serviced after event handler \textit{A} will all be executed after \textit{A}, but are not guaranteed to be executed in any particular order amongst themselves.
+Users can specify the callback order of a handler within its category at the time of registration. 
+Users can specify that a given handler be executed before or after another target handler should both handlers appear in an event chain (the ordering is ignored if the other handler isn't included).
```
jjhursey, line 47: Per 3Q 2023 ASC meeting:
```suggestion
Users can specify that a given handler be executed before or after another target handler should both handlers appear in the event chain (the ordering is ignored if the other handler isn't included). 
```

Chap_API_Event.tex
```diff
@@ -143,10 +146,10 @@ \subsection{\code{PMIx_Register_event_handler}}
...
+Register an event handler to report events. Note that the codes being registered do \textit{not} need to be \ac{PMIx} event constants --- any integer value can be registered. This allows for registration of non-PMIx events such as those defined by a particular \ac{SMS} vendor or by an application itself.
 
 \adviceuserstart
-In order to avoid potential conflicts, users are advised to only define codes that lie outside the range of the \ac{PMIx} standard's error codes. Thus, \ac{SMS} vendors and application developers should constrain their definitions to positive values or negative values beyond the \refconst{PMIX_EXTERNAL_ERR_BASE} boundary.
+In order to avoid potential conflicts, users are advised to only use values as event codes if they lie outside the range of the \ac{PMIx} standard's error and event codes (See \ref{api:struct:usererrors}). Thus, \ac{SMS} vendors and application developers should constrain their definitions to positive values or negative values beyond the \refconst{PMIX_EXTERNAL_ERR_BASE} boundary.
```
abouteiller, line 152:
```suggestion
In order to avoid potential conflicts, users are advised to only use event code values that lie outside the range of the \ac{PMIx} standard's error and event codes (See \ref{api:struct:usererrors}). Thus, \ac{SMS} vendors and application developers should constrain their definitions to positive values or negative values beyond the \refconst{PMIX_EXTERNAL_ERR_BASE} boundary.
```

Chap_API_Event.tex
```diff
@@ -473,7 +476,7 @@ \subsection{\code{PMIx_Notify_event}}
 \returnend
 
 \reqattrstart
-The following attributes are required to be supported by all \ac{PMIx} libraries:
+The following attributes are required to be supported by all \ac{PMIx} libraries and will be passed to all invoked handlers:
```
dsolt, line 480: add a "when relevant" to this intro

jjhursey:
PMIx ASC 4Q 2023 Meeting
 * Passed first vote: 9 yes / 0 no / 0 abstain
 * Can be presented for a second vote during the 1Q 2024 meeting

abouteiller:
PMIx ASC 2024Q1 Meeting

Passed first vote: 9 yes / 0 no / 0 abstain
Accepted as provisional (including 52febfb6).

abouteiller (approved):

abouteiller:
I got trigger happy @pmix/asc-release-managers please double check me.

PR #449, opened 2023-04-24
First pass at IAWG pass at pack/unpack chapter

dsolt:
Signed-off-by: David Solt <dsolt@us.ibm.com>


dsolt:
Outstanding issues not addressed yet in the PR:

- Use of calloc/free in macros and in PMIx_data_load, No means to free the data allocated by PMIx_Data_compress and decompress
- “These errors cannot be detected during packing” - not sure what to do with this.  
- Text talks about resetting the unpack_ptr, but there is no macro to do that.
- If more values exist in the buffer than can fit into the memory storage, then the function will unpack what it can fit into that location and return an error code indicating that the buffer was only partially unpacked.
- Use of “undisturbed” in advice to users
- Why PMIx_data_compress does not take a byte object

dsolt:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

Chap_API_Data_Mgmt.tex
```diff
@@ -243,11 +245,12 @@ \subsection{\code{PMIx_Data_unpack}}
 
 
 \begin{arglist}
-\argin{source}{Pointer to a \refstruct{pmix_proc_t} structure containing the nspace/rank of the process that packed the provided buffer. A NULL value may be used to indicate that the source is based on the same \ac{PMIx} version as the caller. Note that only the source's nspace is relevant. (handle)}
+\argin{source}{Pointer to a \refstruct{pmix_proc_t} structure containing the description of the process that packed the provided buffer. A NULL value may be used to indicate that the source is based on the same \ac{PMIx} version as the caller.
```
rhc54, line 248: Just a little confusing terminology - elsewhere, we refer to a `pmix_proc_t` as being the "process identifier", not a "description of the process". A description of the process sounds like a much larger set of information - e.g., where it is located, what executable is involved, etc.

Chap_API_Data_Mgmt.tex
```diff
@@ -261,23 +264,22 @@ \subsection{\code{PMIx_Data_unpack}}
...
-
-Note that it is possible for the buffer to be corrupted and that \ac{PMIx} will \textit{think} there is a proper variable type at the beginning of an unpack region --- but that the value is bogus (e.g., just a byte field in a string array that so happens to have a value that matches the specified data type flag). Therefore, the data type error check is \textit{not} completely safe.
+The unpack function unpacks the next value (or values) from the given buffer. Providing an unsupported type flag 
+or specifying a data type that \textit{does not} match the type of the next item in the buffer will be reported as an error.
```
rhc54, line 268: Note that this is _only_ true if the library was told to use fully described buffers. Otherwise, we have no idea what the actual type of the data is - we just unpack the bits and put them in the data object of the type you claimed they should be.

Chap_API_Data_Mgmt.tex
```diff
@@ -261,23 +264,22 @@ \subsection{\code{PMIx_Data_unpack}}
...
-
-Note that it is possible for the buffer to be corrupted and that \ac{PMIx} will \textit{think} there is a proper variable type at the beginning of an unpack region --- but that the value is bogus (e.g., just a byte field in a string array that so happens to have a value that matches the specified data type flag). Therefore, the data type error check is \textit{not} completely safe.
+The unpack function unpacks the next value (or values) from the given buffer. Providing an unsupported type flag 
+or specifying a data type that \textit{does not} match the type of the next item in the buffer will be reported as an error. 
+An attempt to read an uninitialized buffer or read beyond the end of the stored data held in the buffer will also return an error.
```
rhc54, line 269: Again, the uninitialized buffer clause isn't really correct - we have no idea if the buffer was initialized or just contains random garbage.

Chap_API_Data_Mgmt.tex
```diff
@@ -261,23 +264,22 @@ \subsection{\code{PMIx_Data_unpack}}
...
 
-Note that any data that was not hard type cast (i.e., not type cast to a specific size) when packed may lose precision when unpacked by a non-homogeneous recipient. \ac{PMIx} will do its best to deal with heterogeneity issues between the packer and unpacker in such cases. Sending a number larger than can be handled by the recipient will return an error code generated upon unpacking --- these errors cannot be detected during packing.
+Note that any data that is packed using a type that does not explicitly specifies its size may lose precision when unpacked by a non-homogeneous recipient. \ac{PMIx} will do its best to deal with heterogeneity issues between the packer and unpacker in such cases. Sending a number larger than can be handled by the recipient will return an error code generated upon unpacking --- these errors cannot be detected during packing.
```
rhc54, line 275: Believe you modified this above to replace "number" with "value"?

Chap_API_Data_Mgmt.tex
```diff
@@ -243,11 +245,12 @@ \subsection{\code{PMIx_Data_unpack}}
 
 
 \begin{arglist}
```
dsolt, line 247: "the description" -> the \refstruct{pmix_proc_t}

Chap_API_Data_Mgmt.tex
```diff
@@ -243,11 +245,12 @@ \subsection{\code{PMIx_Data_unpack}}
...
-\argin{source}{Pointer to a \refstruct{pmix_proc_t} structure containing the nspace/rank of the process that packed the provided buffer. A NULL value may be used to indicate that the source is based on the same \ac{PMIx} version as the caller. Note that only the source's nspace is relevant. (handle)}
+\argin{source}{Pointer to a \refstruct{pmix_proc_t} structure containing the description of the process that packed the provided buffer. A NULL value may be used to indicate that the source is based on the same \ac{PMIx} version as the caller. 
+i. The rank may be set to \refconst{PMIX_RANK_WILDCARD} as only the namespace is used to determine the packing version as all processes in a namespace are required to use the same \ac{PMIx} version.  (handle)}
```
dsolt, line 249: Only the namespace is used..... (get rid of "The rank may be set to ... as"

Chap_API_Data_Mgmt.tex
```diff
@@ -243,11 +245,12 @@ \subsection{\code{PMIx_Data_unpack}}
...
-\arginout{max_num_values}{The number of values to be unpacked --- upon completion, the parameter will be set to the actual number of values unpacked. In most cases, this should match the maximum number provided in the parameters --- but in no case will it exceed the value of this parameter.  Note that unpacking fewer values than are actually available will leave the buffer in an unpackable state --- the function will return an error code to warn of this condition.(\code{int32_t})}
-\argin{type}{The type of the data to be unpacked --- must be one of the \ac{PMIx} defined data types (\refstruct{pmix_data_type_t})}
+\arginout{dest}{A pointer to the memory location into which the data is to be stored. Note that these values will be stored contiguously in memory. For strings, this pointer must be to (char**) to provide a means of supporting multiple string operations. The unpack function will allocate memory for each string in the array, but the caller must provide adequate memory for the array of pointers. (\code{void*})}
```
dsolt, line 251: note that the memory allocation mentioned for strings is really true of all PMIx structures

Chap_API_Data_Mgmt.tex
```diff
@@ -261,23 +264,22 @@ \subsection{\code{PMIx_Data_unpack}}
 %%%%
 \descr
 
-The unpack function unpacks the next value (or values) of a specified type from the given buffer. The buffer must have already been initialized via an \refmacro{PMIX_DATA_BUFFER_CREATE} or \refmacro{PMIX_DATA_BUFFER_CONSTRUCT} call (and assumedly filled with some data) --- otherwise, the unpack_value function will return an error. Providing an unsupported type flag will likewise be reported as an error, as will specifying a data type that \textit{does not} match the type of the next item in the buffer. An attempt to read beyond the end of the stored data held in the buffer will also return an error.
-
```
dsolt, line 265: Make a decision on whether we want to describe behavior of debug vs non-debug and whether to mention that type checking is only done for debug.   Either way, don't may claim that type matching is done in all cases.

Chap_API_Data_Mgmt.tex
dsolt, line 275: same text as above for "number larger than" -> value outside the range that ...

Chap_API_Data_Mgmt.tex
```diff
@@ -261,23 +264,22 @@ \subsection{\code{PMIx_Data_unpack}}
...
 The namespace of the process that packed the buffer is used solely to resolve any data type
 differences between \ac{PMIx} versions. The packer must, therefore, be
-known to the user prior to calling the pack function so that the
-\ac{PMIx} library is aware of the version the packer is using. Note that
+known to the caller prior to calling the unpack function so that the
+\ac{PMIx} library is aware of the version the packer used. Note that
 all processes in a given namespace are \textit{required} to use the same \ac{PMIx}
-version --- thus, the caller must only know at least one process from the
-packer's namespace.
+version --- thus the target rank may be set to \refconst{PMIX_RANK_WILDCARD}.
```
dsolt, line 283: back out this change based on decision to not claim that rank might be set to WILDCARD

Chap_API_Data_Mgmt.tex
```diff
@@ -337,7 +339,7 @@ \subsection{\code{PMIx_Data_print}}
 }
 
 \begin{arglist}
-\argin{output}{The address of a pointer into which the address of the resulting output is to be stored. (\code{char**})}
+\argout{output}{The address of a pointer into which the address of the resulting output is to be stored. (\code{char**})}
 \argin{prefix}{String to be prepended to the resulting output (\code{const char*})}
```
dsolt, line 343: double check similar API's to see if they use argin or argout for this

Chap_API_Data_Mgmt.tex
```diff
@@ -352,7 +354,7 @@ \subsection{\code{PMIx_Data_print}}
 %%%%
 \descr
 
-Since registered data types can be complex structures, the system needs some way to know how to print them (i.e., convert them to a string representation). Primarily for debug purposes.
+Since registered data types can be complex structures, the \ac{PMIx} provides \refapi{PMIx_Data_print} to print them (i.e., convert them to a string representation). Primarily for debug purposes.  Note that the format of the resulting string may vary from one implementation to another.
```
dsolt, line 357: extra the in "the \ac{PMIx}"

Chap_API_Data_Mgmt.tex
```diff
@@ -562,9 +565,9 @@ \subsection{\code{PMIx_Data_decompress}}
 \cspecificend
 
 \begin{arglist}
-\argout{outbytes}{Address where the pointer to the decompressed data region is to be returned (handle)}
-\argout{nbytes}{Address where the number of bytes in the decompressed data region is to be returned (handle)}
 \argin{inbytes}{Pointer to the source data (handle)}
+\argout{nbytes}{Address where the number of bytes in the decompressed data region is to be returned (handle)}
+\argout{outbytes}{Address where the pointer to the decompressed data region is to be returned (handle)}
 \argin{size}{Number of bytes in the source data region (\code{size_t})}
```
dsolt, line 571: Looks like nybtes and size are still reversed in order.

Chap_API_Data_Mgmt.tex
```diff
@@ -199,14 +199,16 @@ \subsection{\code{PMIx_Data_pack}}
 
 The pack function packs one or more values of a specified type into the specified buffer.  The buffer must have already been
 initialized via the \refmacro{PMIX_DATA_BUFFER_CREATE} or \refmacro{PMIX_DATA_BUFFER_CONSTRUCT}
-macros --- otherwise, \refapi{PMIx_Data_pack} will return an error.
+macros --- otherwise, \refapi{PMIx_Data_pack} will return an error.  
+The buffer may have already contain packed data, in which case the new data is appended to the buffer.
 Providing an unsupported type flag will likewise be reported as an error.
 
-Note that any data to be packed that is not hard type cast (i.e.,
-not type cast to a specific size) may lose precision when unpacked
```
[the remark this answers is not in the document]
dsolt: also changed on line approx 275:

Chap_API_Data_Mgmt.tex
dsolt: Not changing this as described above (which would result in "Pointer to a \refstruct{pmix_proc_t} structure containing the \refstruct{pmix_prot_t}").  We think this double occurence of pmix_proc_t is still confusing so went with "Pointer to a \refstruct{pmix_proc_t} structure of the process that packed the provided buffer."

Chap_API_Data_Mgmt.tex
dsolt: Not sure what to do with this.  We didn't change anything around this and not sure how to fix it.

rhc54: Suggestion: "For objects/structures, this point must be to a double-star to provide a means of supporting multiple object/structure operations. The unpack function will allocate memory for each object/structure in the array, but the caller must provide adequate memory for the array of pointers"

Chap_API_Data_Mgmt.tex
dsolt: This change looks good to me after review.  Consistent with other uses of argout

Chap_API_Data_Mgmt.tex
dsolt: We have gone in the direction of not describing debug behavior

dsolt:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

abouteiller:
Passed 2nd vote at ASC 24Q4 so it is now accepted.

abouteiller (approved):
Verified to look normal in pdf form on top of master

Issue #354, opened 2021-06-30
Help understanding info's passed to pmix_server_dmodex_req_fn_t


dsolt:
This might have to be a @rhc54 question.   I see that pmix_server_dmodex_req_fn_t is called by the PMIx server library and it has an info and ninfo fields.   It is unclear what a RM should do with these fields.  For example, I'm seeing it called in one case with 2 infos:  pmix.req.key and pmix.node.info.   The RM forwards this request to the appropriate node where the RM calls PMIx_server_dmodex_request(const pmix_proc_t *proc, pmix_dmodex_response_fn_t cbfunc, void *cbdata).  However, there is no mechanism to pass the 2 info's provided into this call.  So I'm not sure how the RM is supposed to handle these.  The standard only calls out pmix.req.key as a required attribute, but it seems like the pmix server passed in pmix.node.info for a reason.  For the pmix.req.key, the RM can't even tell if the data returned has the required key.  It just gets back a "blob" of data, so its not like the RM can loop calling PMIx_server_dmodex_request repeatedly until the required attribute is returned.

Thoughts on how this is supposed to work and then maybe we can work on making the standard present it clearer?

### Suggested Clarification

I don't have a suggestion right now until I have a better idea of how it is supposed to work.

### References

17.3.7
17.2.10.1

rhc54:
Hmmm...it looks like there may be a problem in the implementation, actually, if it is doing that - though I suppose someone might mistakenly call `PMIx_Get` with a non-reserved key and flag it as "node info". In the chapter on "reserved keys", we explicitly state that the server cannot request reserved keys from the host environment. Reserved keys can only be provided by the host during namespace registration (and updated via the same API). The dmodex request function, therefore, is solely for requesting non-reserved keys.

Thus, it makes no sense to pass a dmodex request that contains the `PMIX_NODE_INFO` attribute - the host would have no idea what to do with it as all "node info" are reserved keys, so the only option would be to ignore it (which is certainly what PRRTE would do). The only attributes the dmodex request should be able to support are things like "required key" and "timeout" that must be dealt with by the host (so no reason to pass anything down to the PMIx server library). I know we are checking this in the latest library to ensure that we never pass a reserved key request up to the host, but I honestly forget how far back we went with that change. We don't, however, check the attributes to see if someone mistakenly included "node info" (or one of its colleagues) for a non-reserved key - we just ignore it.

[...]

PR #484, opened 2023-05-11
How to implement pmix_server_dmodex_req_fn_t when PMIX_REQUIRED_KEY set

dsolt:


dsolt:
Issue #354

Chap_API_Server.tex
```diff
@@ -2411,18 +2411,19 @@ \subsection{\code{pmix_server_dmodex_req_fn_t}}
...
 
+
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsubsection{Dmodex attributes}
 
 %
 \declareAttribute{PMIX_REQUIRED_KEY}{"pmix.req.key"}{char*}{
-Identifies a key that must be included in the requested information. If the specified key is not already available, then the \ac{PMIx} servers are required to delay response to the dmodex request until either the key becomes available or the request times out.
+Identifies a key that must be included in the requested information. If the specified key is not already available, then the \ac{PMIx} servers are required to delay response to the dmodex request until either the key becomes available or the request times out.  The server can test the availability of a required key by calling \refapi{PMIx_Get} on the key.  Since there is no mechanism to retract a key-value that has been made available by a client through PMIx_Put, a successful \refapi{PMIx_Get} call on the key by the server is sufficient to ensure that a subsquent \refapi{PMIx_server_dmodex_request} call targeting the process will include the key.
```
abouteiller, line 2426:
```suggestion
Identifies a key that must be included in the requested information. If the specified key is not already available, then the \ac{PMIx} servers are required to delay response to the dmodex request until either the key becomes available or the request times out.  The server can test the availability of a required key by calling \refapi{PMIx_Get} on the key.  Since there is no mechanism to retract a key-value that has been made available by a client through PMIx_Put, a successful \refapi{PMIx_Get} call on the key by the server is sufficient to ensure that a subsequent \refapi{PMIx_server_dmodex_request} call targeting the process will include the key.
```

I think I'd like this to be an advise, given that it hints at a possible implementation without necessarily mandating it.

dsolt:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

jjhursey:
PMIx ASC 4Q 2023 Meeting
 * Passed first vote: 9 yes / 0 no / 0 abstain
 * Can be presented for a second vote during the 1Q 2024 meeting

abouteiller:
PMIx ASC 2024Q1 Meeting

Passed first vote: 9 yes / 0 no / 0 abstain
Accepted as provisional.

abouteiller (approved):

abouteiller:
@dsolt @raffenet ready to merge

PR #493, opened 2023-09-06
Remove duplicate PMIx_Get description

raffenet:


raffenet:
As discussed in the September 2023 ASC monthly meeting, this is likely an editorial change and could be merged outside of the reading/voting process.

jjhursey (approved):

jjhursey:
If @abouteiller also approves and agrees, I think this can be merged in without formal reading per Ken's note above. This seems to be purely editorial.

abouteiller (approved):

abouteiller:
does it need back ported to v4.x?

Issue #491, opened 2023-08-10
Tickets merged into 4.2 need to be mirrored into 5.1


abouteiller:
All tickets that are merged into 4.2 also need to be mirrored into 5.1

https://github.com/pmix/pmix-standard/issues?q=is%3Aissue+milestone%3A%22PMIx+v4.2+Standard%22

* [x] All merged PRs reviewed that they have been replicated into v5
* [x] All open 4.2 PRs have a v5 replica

abouteiller:
#495 still pending resolution

abouteiller:
Went over individual commits from `git merge-base master v4`  `95556c07c7f3569ea014724b9b74d8ff66087fb7` and found nothing missing from v4 to v5

PR #488, opened 2023-06-07
Change type of key parameter to PMIx_Get_nb to match PMIx_get

abouteiller:
Change type of key parameter to PMIx_Get_nb to match PMIx_get (i.e., pmix_key_t)

PR #495, opened 2023-09-14
Issue420/get nb key@v5

### LINKED ISSUE #420: Parameter Conflict: PMIx_Get vs PMIx_Get_nb
PMIx_Get has a signature that includes `const pmix_key_t key`.

PMIx_Get_nb has a signature that includes `const char key[]` for the same argument.

The problem is that these conflict in the eyes of today's picky compilers such as gcc12, which will generate a warning if:

* you pass a `char*` string to PMIx_Get, but not if you pass that same string to PMIx_Get_nb.
* you pass a `pmix_key_t` to PMIx_Get_nb, but not if you pass that same key to PMIx_Get

So which is it supposed to be?

abouteiller:
Import changes from v4.2 to v5.1

#491
closes #420 

* [x] waiting on #488 merged in v4
* [x] review openPMIx for conformity with proposed change: NOT CONFORMANT
* [x] review v4.2 and potential need to rollback #488 ? NEED TO ROLLBACK


raffenet (dismissed):

raffenet:
OK, after I clicked approve I went and double checked the OpenPMIx and see that neither API there uses `pmix_key_t`. I forget if this was discussed when we proposed the change?
```c
PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc, const char key[],
                                   const pmix_info_t info[], size_t ninfo,
                                   pmix_value_t **val);
PMIX_EXPORT pmix_status_t PMIx_Get_nb(const pmix_proc_t *proc, const char key[],
                                      const pmix_info_t info[], size_t ninfo,
                                      pmix_value_cbfunc_t cbfunc, void *cbdata);
```

rhc54:
Yeah, you can't make it a `pmix_key_t` as the `key` parameter can be `NULL` - and IIRC, the compiler will complain if it is declared as a `pmix_key_t` in that instance.

abouteiller:
We need to double check what happened with #488 as this is when we voted on it. The way that was understood (see #420) the implementation was already using the pmix_key_t and one of the procedures took it as such.

rhc54:
> implementation was already using the pmix_key_t and one of the procedures took it as such.

Correct - and as I stated in #420, we corrected it to both APIs using `const char key[]`. Otherwise, you cannot pass a `NULL` argument, which would break some of the rules for `PMIx_Get`

abouteiller:
I did exactly the wrong fix... Good that we caught that before it went out the door. I'll update #420 and we will have to revote on it, I'll check if that'll delay v4.2 based on our rules but I think it will.

abouteiller:
This will now need to mirror #510

raffenet (approved):

PR #500, opened 2023-10-02
Initial commit for IAWG review of Process Management chapter

dsolt:


Chap_API_Proc_Mgmt.tex
```diff
@@ -239,7 +217,7 @@ \subsection{\code{PMIx_Spawn_nb}}
 \reqattrend
 
 \optattrstart
-The following attributes are optional for host environments that support this operation:
+The following attributes are optional for \ac{PMIx} implementations:
```
rhc54, line 220: Hmmm...those operations are not performed within PMIx. The attribute is simply communicated to the host for execution. I suspect you are trying to say that a PMIx implementation doesn't have to define them, but that doesn't make much sense as we don't provide that option elsewhere. You have to define it - the host just may not implement support for it.

Chap_API_Proc_Mgmt.tex
```diff
@@ -356,15 +372,15 @@ \subsection{Spawn attributes}
 }
 %
 \declareAttribute{PMIX_HOSTFILE}{"pmix.hostfile"}{char*}{
-Hostfile to use for spawned processes.
+Hostfile to use for spawned processes.  The format of this file is determined by the \ac{PMIx} implementation or host environment.  The file may not be portable across implementations.
```
rhc54, line 375: Nothing to do with the PMIx implementation - we just pass the name of the file.

Chap_API_Proc_Mgmt.tex
```diff
@@ -356,15 +372,15 @@ \subsection{Spawn attributes}
...
 }
 %
 \declareAttribute{PMIX_ADD_HOST}{"pmix.addhost"}{char*}{
 Comma-delimited list of hosts to add to the allocation.
 }
 %
 \declareAttribute{PMIX_ADD_HOSTFILE}{"pmix.addhostfile"}{char*}{
-Hostfile containing hosts to add to existing allocation.
+Hostfile containing hosts to add to existing allocation.  The format of this file is determined by the \ac{PMIx} implementation or host environment.  The file may not be portable across implementations.
```
rhc54, line 383: Ditto here - we only pass the name. We do not process the file.

Chap_API_Proc_Mgmt.tex
```diff
@@ -376,7 +392,7 @@ \subsection{Spawn attributes}
 }
 %
 \declareAttribute{PMIX_DISPLAY_MAP}{"pmix.dispmap"}{bool}{
-Display process mapping upon spawn.
+Provide process mapping upon spawn.  Format of the provided map is implementation specific.
```
rhc54, line 395: Ouch - no, we don't provide process mapping. We simply pass the attribute to the host requesting that the _host_ display the resulting process map. We don't do anything.

Chap_API_Proc_Mgmt.tex
```diff
@@ -432,7 +448,7 @@ \subsection{Spawn attributes}
 }
 %
 \declareAttribute{PMIX_INDEX_ARGV}{"pmix.indxargv"}{bool}{
-Mark the \code{argv} with the rank of the process.
+If set to true, will use the given name of the executable (\code{argv[0]}) as a base name and each rank will be invoked using the base name with the string "-<\emph{rank}>" appended to it, where \emph{rank} is the \ac{PMIx} rank of the process being invoked.
```
rhc54, line 451: Should probably make clear that PMIx isn't doing this - we are passing an attribute to the host environment requesting that _it_ do this. Whether or not it happens is completely up to the host.

Chap_API_Proc_Mgmt.tex
```diff
@@ -448,7 +464,7 @@ \subsection{Spawn attributes}
 }
 %
 \declareAttribute{PMIX_REPORT_BINDINGS}{"pmix.repbind"}{bool}{
-Report bindings of the individual processes.
+Report bindings of the individual processes.  How and where this information is reported is implementation dependent as well as dependent on whether the processes are created through a launching tool or by a direct call to \refapi{PMIx_Spawn}.
```
rhc54, line 467: Again, PMIx isn't doing anything here - we just pass the attribute to the host requesting that _it_ generate this report.

Chap_API_Proc_Mgmt.tex
```diff
@@ -468,7 +484,7 @@ \subsection{Spawn attributes}
 }
 %
 \declareAttribute{PMIX_SPAWN_TOOL}{"pmix.spwn.tool"}{bool}{
-Indicate that the job being spawned is a tool.
+Indicate that the job being spawned is a tool.  The repercussions of setting this attribute varies based on implementation.  For example, some implementations may not perform cpu-binding on process marked as a tool.
```
rhc54, line 487: PMIx implementation isn't doing anything here - we just pass the attribute informing the host that the spawn request is launching a tool. What that means is totally up to the host.

rhc54 (commented):
Hope the comments help clarify things.

Chap_API_Proc_Mgmt.tex
dsolt: Thanks for the feedback.  We will re-visit these and consider updating to make it clearer that PMIx is just the messenger in all of these use cases.   We just wanted to make it clear that the format or input/output was not designated by PMIx, but agree that the best way to do that is to make it cleat that PMIx is only passing these through.

Chap_API_Proc_Mgmt.tex
dsolt: Suggestion during working group:   Request that the host environment provide process mapping information about the spawn operation.

dsolt: Our motivation for changing this from "display" to "provide" is that we didn't want to specify that the anything (be it pmix itself or host environment) has to display the mapping to a terminal.   Display seems to preclude options such as writing to a file or many other means it might use to make this information available to the user.

Chap_API_Proc_Mgmt.tex
dsolt: So we generally have a mix of the following:

dsolt: This chapter even has sections where some attributes are specifically called out as optional for PMIx implementations and other attributes are marked as optional for host environments for the same API.

Chap_API_Proc_Mgmt.tex
dsolt: This comes down to whether we want to force the division of labor between PMIx or the host environment.  We didn't want to preclude the possibility that someone say, took OpenPMix and decided to implement spawn directly inside of it.   If we think that is never going to happen, we can go with:

The format of this file is determined by the host environment.  A hostfile may not be portable across systems with different host environments.

Chap_API_Proc_Mgmt.tex
dsolt: What about:

Provide process mapping upon spawn.  The format of the provided map is specific to the host environment providing it.

Chap_API_Proc_Mgmt.tex
rhc54: Sure - it is optional to implement support for those attributes. We require support for others, and we are very clear which ones are optional for the implementation vs the host vs both.

It is critical that PMIx remain in the "messenger" role and not shift into being the "doer" of things. We do internally support PMIx functions, but we want to be very careful and clear where the line lies. And no, that isn't optional across PMIx implementations - or else the host now has to deal with uncertainty over what the implementation may or may not do, which will be very unpopular (should there ever be another implementation).

Chap_API_Proc_Mgmt.tex
rhc54: We do support local fork/exec from within the PMIx library, but that is nowhere equivalent to spawn as we provide no support for launching things across nodes. We really should stick with the divisions we have had since day one and not muddy the water.

Chap_API_Proc_Mgmt.tex
rhc54: We should not imply that PMIx is going to provide the mapping - we don't do mapping, so it makes no sense for us to try and display/provide it. This is the responsibility of the host as they actually did the mapping. Format of the map is certainly host environment specific, and we don't specify it anywhere.

Chap_API_Proc_Mgmt.tex
rhc54: Just noting that the change adds language like "...will use...", but leaves hanging the question of "who" will use the name. If you want to clarify, it might be something like "host is requested to use...". Or you can split this list to indicate which attributes are passed to the host for execution. Not entirely sure of the best way. Just trying to help you in your quest to clarify exactly what we mean here, and not replace one confusion with another.

Chap_API_Proc_Mgmt.tex
rhc54: I think the question here again is that you stated the behavior "...varies based on implementation", but not on which implementation - the host or the PMIx library? I'm not aware of any PMIx repercussions, nor was that the intent of the attribute - it was provided so that the host environment would know this is a tool being spawned, and _if_ they wanted to do something different for tools, then they could do so.

rhc54 (commented):
Just trying to help clarify - I think much of this has to do with clarifying which implementation (host or PMIx) your changes are talking about.

dsolt:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

Chap_API_Proc_Mgmt.tex
```diff
@@ -376,7 +394,7 @@ \subsection{Spawn attributes}
 }
 %
 \declareAttribute{PMIX_DISPLAY_MAP}{"pmix.dispmap"}{bool}{
-Display process mapping upon spawn.
+Provide process mapping upon spawn.  The format of the provided map is specific to the host environment providing it.
```
abouteiller, line 397:
```suggestion
Display process mapping upon spawn.  The format of the displayed map is specific to the host environment providing it.
```
should we specify if the map will appear on stderr/stdout?

Chap_API_Proc_Mgmt.tex
```diff
@@ -1505,7 +1560,7 @@ \subsection{Device type}
 The \refstruct{pmix_device_type_t} is a \code{uint64_t} bitmask for identifying the type(s) whose distances are being requested, or the type of a specific device being referenced (e.g., in a \refstruct{pmix_device_distance_t} object).
 
 \copySignature{pmix_device_type_t}{1.0}{
-typedef uint16_t pmix_device_type_t;
+typedef uint64_t pmix_device_type_t;
```
abouteiller, line 1563: is this an ABI change?

Chap_API_Proc_Mgmt.tex
```diff
@@ -806,21 +851,19 @@ \subsection{\code{PMIx_Connect}}
 \reqattrend
 
 \optattrstart
-The following attributes are optional for \ac{PMIx} implementations:
-
-\pasteAttributeItem{PMIX_ALL_CLONES_PARTICIPATE}
-
-
 The following attributes are optional for host environments that support this operation:
```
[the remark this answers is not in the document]
dsolt: confirmed that the optattrstart only contains the banner and not the intro text.  Something to consider since we repeat that "The following attributes are optional...." everywhere in the standard, but that is an issue for another PR.

Chap_API_Proc_Mgmt.tex
```diff
@@ -1313,6 +1354,7 @@ \subsection{\code{PMIx_Parse_cpuset_string}}
 }
 
 \begin{arglist}
+\argin{cpuset_string}{\refattr{PMIX_CPUSET} string associated with a \ac{PMIx} process}
 \argin{cpuset_string}{String returned by the \refapi{PMIx_server_generate_cpuset_string} \ac{API} (handle)}
 \arginout{cpuset}{Address of an object where the bitmap is to be stored (memory reference)}
```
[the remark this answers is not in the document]
dsolt: Fixed in a new commit

Chap_API_Proc_Mgmt.tex
```diff
@@ -219,11 +187,21 @@ \subsection{\code{PMIx_Spawn_nb}}
 
 \returnsimplenb
 
-\returnstart
+If executed, the status returned in the provided callback function will be one of the following constants:
+
 \begin{itemize}
-    \item \refconst{PMIX_OPERATION_SUCCEEDED}, indicating that the request was immediately processed and returned \textit{success} - the \refarg{cbfunc} will \textit{not} be called
+\item \refconst{PMIX_SUCCESS} All data was found and has been returned.
+\item \refconst{PMIX_ERR_JOB_ALLOC_FAILED} The job request could not be executed due to failure to obtain the specified allocation
```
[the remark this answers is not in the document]
dsolt: Changed to "The operation was successfully completed." which is what is used in at least one other place.

Chap_API_Proc_Mgmt.tex
dsolt: Added to the upcoming commit with changes from reading

Chap_API_Proc_Mgmt.tex
```diff
@@ -166,20 +120,34 @@ \subsection{\code{PMIx_Spawn}}
...
-information so it can query job-level info without incurring any communication penalties, (b) newly spawned child processes will receive a copy of the parent processes job-level info, and (c) both the parent process and members of the child job will receive notification of errors from processes in their combined assemblage.
+for details).  
+Both the parent process and members of the child job will receive notification of errors from processes in their combined assemblage.
+
+\adviceimplstart
+It is recommended that an implementation will cause the parent process to be given a copy of the new job's
+information so it can query job-level info without incurring any communication penalties.  Similarly, the newly 
+spawned child processes should receive a copy of the parent processes' job-level info due to the high likelihood 
+that the child will make subsequent queries about its parent.
+\adviceimplend
```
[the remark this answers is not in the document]
dsolt: changed to advicermstart.   Clarified "... a copy of the new job's *job-level* information..."

Chap_API_Proc_Mgmt.tex
```diff
@@ -219,11 +187,21 @@ \subsection{\code{PMIx_Spawn_nb}}
...
+\item \refconst{PMIX_ERR_JOB_APP_NOT_EXECUTABLE} The specified application executable either could not be found, or lacks execution privileges.
+\item \refconst{PMIX_ERR_JOB_NO_EXE_SPECIFIED} The job request did not specify an executable.
+\item \refconst{PMIX_ERR_JOB_FAILED_TO_MAP} The launcher was unable to map the processes for the specified job request.
+\item \refconst{PMIX_ERR_JOB_FAILED_TO_LAUNCH} One or more processes in the job request failed to launch
+\item \refconst{PMIX_ERR_JOB_EXE_NOT_FOUND} Specified executable not found
+\item \refconst{PMIX_ERR_JOB_INSUFFICIENT_RESOURCES} Insufficient resources to spawn job
+\item \refconst{PMIX_ERR_JOB_SYS_OP_FAILED} System library operation failed
+\item \refconst{PMIX_ERR_JOB_WDIR_NOT_FOUND} Specified working directory not found
+\item a non-zero \ac{PMIx} error constant indicating a reason for the request's failure.
 \end{itemize}
```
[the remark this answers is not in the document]
dsolt: We do not appear to have one.  Looks like one should be pretty easy to create, but should be applied across all chapters?

Chap_API_Proc_Mgmt.tex
abouteiller: this is an errata, the ABI document should already have it listed as a uint64_t

Chap_API_Proc_Mgmt.tex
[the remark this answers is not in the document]
abouteiller: tracking issue https://github.com/pmix/pmix-standard/issues/512

naughtont3:
PR500 passed the 2nd vote at ASC 2024-Q3 meeting.

abouteiller (approved):

PR #501, opened 2023-10-17
Bump version to 5.1 and add release script

raffenet:
Version 5.0 was tagged and released in May 2023. The master branch is now tracking version 5.1. We neglected to update the revision history chapter with the release date, so fix that and add a stub for 5.1 and use `\VERDATE` to substitute the release date automatically.

Add a new make target `make release`. This invokes `bin/release.sh` to automatically remove the watermark and substitute the current month and year for the release date. This process can be extended to support other release checklist type items.

### DIFF
--- a/Chap_Revisions.tex
+++ b/Chap_Revisions.tex
@@ -1304,7 +1304,7 @@ Buffer.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %%%%%%%%%% History: Version 5.0
-\section{Version 5.0: TBD}
+\section{Version 5.0: May 2023}
 
 The v5.0 update includes the following changes from the v4.1 document:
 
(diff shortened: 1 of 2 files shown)


abouteiller (dismissed):

Makefile
```diff
@@ -79,6 +79,11 @@ pmix-standard.pdf: $(CHAPTERS) $(SOURCES) pmix.sty pmix-standard.tex figs/pmix-l
 	$(LATEX_C) -interaction=batchmode pmix-standard.tex
 	@cp pmix-standard.pdf pmix-standard-${version}.pdf
 
+release:
+	@./bin/release.sh set
+	@make all
+	@./bin/release.sh restore
```
raffenet, line 85: I'm going to make a small change to this create the release in a temporary directory instead of the working copy. This way if something goes wrong, it won't pollute the state of your repo. This is the way the MPICH release script works.

raffenet:
Removing the release script from this PR so we can just update the doc and unblock other PRs.

abouteiller (approved):

Issue #508, opened 2024-04-11
PMIX v4.2 Checklist


abouteiller:
### Overview

This is a general issue for tracking progress towards v4.2 completion

### Discussion Items

* v4.2rc3 draft document: 
[pmix-standard-4.2rc3.pdf](https://github.com/pmix/pmix-standard/files/15172311/pmix-standard-4.2rc3.pdf)
[pmix-standard-4.2rc2.pdf](https://github.com/pmix/pmix-standard/files/14943447/pmix-standard-4.2rc2.pdf)

### Prepare final document

* [x] #488 errata is wrong and does the backward change compared to what we actually needed to do (issue #420). We need to correct this before we release. #510 needs merged.
* [x] double check that revision history is complete
* [x] Reference v5.0 in history
* [x] Add acknowledgement section
* [x] Change VER and VERDATE in the main tex file
* [x] Change release date (TBD) in Chap_revisions.tex
* [x] Change copyright date on front matter page (maybe use verdate)

abouteiller:
v4.2 is now published 🥳

PR #509, opened 2024-04-11
V4.2rc2

abouteiller:
Final prep changes for v4.2rc2

Tracking issue #508


abouteiller:
I pushed the changes directly instead of doing a PR

Issue #513, opened 2024-06-13
Import v4.2 changelog to v5.1 (master)


abouteiller:
### Overview

v4.2 has been published, we need to import the changelog to v5.1/master

PR #514, opened 2024-06-13
import v4.2 changelog; populate v5.1 changelog with already voted-in items

abouteiller:
Update master revision history
* import v4.2 changelog
* populate v5.1 changelog with already voted in items
* [x] special case these in the python the double deprecated constants in v4.2 and v5.1
closes:  #513

### DIFF
--- a/Chap_Revisions.tex
+++ b/Chap_Revisions.tex
@@ -1457,14 +1457,155 @@ Replaced by \refconst{PMIX_ERR_PARTIAL_SUCCESS}.
 \end{constantdesc}
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%%%%%%%%%% History: Version 4.2
+\section{Version 4.2: May 2024}
+
+The v4.2 update released after v5.0 to backport select features to the v4 line.
+The v4.2 update succeeds v4.1; changes made in v5.0 do not apply to v4.2
+except if listed again in this changelog section.
+
+The v4.2 update includes the following changes from the v4.1 document:
+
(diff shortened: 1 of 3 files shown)


abouteiller:
We have an unexpected difficulty: we deprecated some attributes and constants twice (in 5.0, and again in 4.2). This causes our automated checker to fail (because it verifies that we indeed deprecate and remove things only once, see `check-attr-refs.py` 

We may have to special-case the checker to not complain about these instances :(

abouteiller:
Solution preferred during discussion at ASC 24M06 is to special case these in the python.

naughtont3:
Errata PR 514 vote passed from the ASC 2024-Q3 voting

abouteiller:
Should be ready to go please double check my python

Chap_Revisions.tex
```diff
@@ -1456,23 +1456,167 @@ \subsection{Removed Constants}
...
+\pasteAttributeItem{PMIX_IOF_LOCAL_OUTPUT}
+\pasteAttributeItem{PMIX_IOF_MERGE_STDERR_STDOUT}
+\pasteAttributeItem{PMIX_IOF_RANK_OUTPUT}
+\pasteAttributeItem{PMIX_IOF_OUTPUT_RAW}
+\pasteAttributeItem{PMIX_IOF_OUTPUT_TO_DIRECTORY}
+\pasteAttributeItem{PMIX_IOF_OUTPUT_TO_FILE}
+
+\subsection{Deprecated constants}
+
+The following constants were deprecated in v4.2 (backport from v5.0):
```
abouteiller, line 1558: move the back port to the section title when all are back ports

PR #522, opened 2025-06-08
Replace the "log" server module fn with a "log2"

rhc54:
The current "log" server module function was defined with a void return. This contrasts with most of the other module functions which characteristically return a pmix_status_t, thereby supporting a return status that indicates if the function is not supported or has been atomically completed.

Add a replacement "log2" function that is identical in signature except for returning a pmix_status_t instead of void. Deprecate the current "log" function.


rhc54:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

rhc54:
@neicker Does this look okay to you?

neicker:
> @neicker Does this look okay to you?

Looks good to me.

rhc54:
I'll try to address those points here - thx!

rhc54:
@neicker I have updated the text to address your questions - please see if this is adequate or further clarification is required.

naughtont3:
Note from [25Q3 meeting](https://github.com/pmix/pmix-standard/wiki/ASC-Q3-2025-Meeting), will allow more time for reviews/comments and bring for vote at next quarterly (25Q4).

Chap_API_Server.tex
```diff
@@ -3333,6 +3334,62 @@ \subsection{\code{pmix_server_tool_connection2_fn_t}}
...
+\pasteAttributeItem{PMIX_VERSION_INFO}
+
+\optattrend
+
+%%%%
+\descr
+
+This function module entry has been \textbf{DEPRECATED} in favor of \refapi{pmix_server_tool_connection2_fn_t}
+
+
```
abouteiller, line 3392: This is a mop-up/cleanup from a prior PR that introduced connection2?

Chap_API_Job_Mgmt.tex
```diff
@@ -886,48 +886,59 @@ \subsection{\code{PMIx_Log}}
...
 
 \pasteAttributeItem{PMIX_LOG_SOURCE}
 \pasteAttributeItem{PMIX_LOG_TIMESTAMP}
 \pasteAttributeItem{PMIX_LOG_GENERATE_TIMESTAMP}
 \pasteAttributeItem{PMIX_LOG_TAG_OUTPUT}
 \pasteAttributeItem{PMIX_LOG_TIMESTAMP_OUTPUT}
 \pasteAttributeItem{PMIX_LOG_XML_OUTPUT}
-\pasteAttributeItem{PMIX_LOG_EMAIL}
-\pasteAttributeItem{PMIX_LOG_EMAIL_ADDR}
-\pasteAttributeItem{PMIX_LOG_EMAIL_SENDER_ADDR}
```
abouteiller, line 917: A number of these have been removed from this list but not reintroduced in another list, should they be added to the list of optional attributes above?

Chap_API_Server.tex
rhc54: Yes - we have the unfortunate problem of conflicts between all these updates, so the diff once we merge becomes...a bloody mess.

Chap_API_Job_Mgmt.tex
rhc54: `PMIX_LOG_EMAIL` is still in the list, just moved to the section immediately above this one. The other `LOG_EMAIL_xxx` attributes are listed in the `PMIX_LOG_EMAIL` description as they are to be included in the `pmix_data_array_t` for that attribute. We can split them into a list within that description (as was done for the stats area) if that helps.

rhc54:
Rebased without conflicts!

PR #523, opened 2025-06-13
Errata: fix a couple of spots in server upcall functions

rhc54:
Server upcall functions were required to return prior to executing the callback function. This was a limitation of the implementation and not a requirement of the Standard - the statements should never have been included here.

FWIW: the limitation has subsequently been removed from the reference implementation.

In the pmix_server_iof_fn_t, clarify that an error status can be returned either directly by the function or in the callback function. It does not require that the host execute the callback to return an error, as was implied by the current language.

### DIFF
--- a/Chap_API_Event.tex
+++ b/Chap_API_Event.tex
@@ -43,8 +43,8 @@ Event notifications provide the process identifier of the source of the event pl
 %
 \end{itemize}
 
-Users can specify the callback order of a handler within its category at the time of registration. 
-Users can specify that a given handler be executed before or after another target handler should both handlers appear in the event chain (the ordering is ignored if the other handler isn't included). 
+Users can specify the callback order of a handler within its category at the time of registration.
+Users can specify that a given handler be executed before or after another target handler should both handlers appear in the event chain (the ordering is ignored if the other handler isn't included).
 The ordering is dictated by providing the event handler name of the target.  The name must have been assigned when the target handler was registered.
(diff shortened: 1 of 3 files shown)


rhc54:

      we know your current thoughts

      want to be an advocate for it
      other technologies instead

abouteiller:
Note for when merging, we will need to add a changelog errata entry

neicker:
While this seem to cover all such mentions in Server-Specific Interfaces section, I found a few more also in other parts:
- Chap_API_Event.tex:118: "The callback function must not be executed prior to returning from the \ac{API}, and ..."
- Chap_API_Query.tex:191: "Note that the library must not invoke the callback function prior to returning from the \ac{API}."

Furthermore a similar sentence is in one of the commands:
- pmix.sty:900: "\newcommand{\returnsimplenb}{.... Note that the library must not invoke the callback function prior to returning from the \ac{API}. ... .}"
This one is used at several places all over the document including in the Server-Specific Interfaces section.

Unclear to me if we want to handle them in this PR or if this would be a new issue.

rhc54:
> In the Revision History section for version 2.1 it is stated:
> 
>     * Clarify return values for non-blocking APIs and emphasize that callback functions must not be invoked prior to return from the API
> 
> 
> Thus, it seems there was a clear intention at that time. Do we have to explain why we have a different view on this now?

We'll note that we changed it in the revision history. Probably have a little blurb, but nothing too verbose. Just didn't want to include it here as the revision history for v6 is created in a different PR - no need to maximize conflict.

PR #524, opened 2025-06-13
Replace the "tool_connection" fn with a "tool_connection2"

rhc54:
The current "tool_connection" server module function was defined with a void return. This contrasts with most of the other module functions which characteristically return a pmix_status_t, thereby supporting a return status that indicates if the function is not supported or has been atomically completed.

Add a replacement "tool_connection2" function that is identical in signature except for returning a pmix_status_t instead of void. Deprecate the current "tool_connection" function.


naughtont3:
2025 ASC Q3 Voting Result: Passed (9 Yes, 0 No, 0 Abstain)

